PR #2, opened 2023-11-03
adds its mockup for sui-native assets

damirka:


Foivos:
ITS in in main at this point.

PR #21, opened 2024-04-19
fix(scripts): after publishing tx, always return an hex of length 64 chars

jcs47:


scripts/publish-package.js
```diff
@@ -51,10 +51,13 @@ async function publishPackage(packagePath, client, keypair) {
 	});
 	if(publishTxn.effects?.status.status != 'success') throw new Error('Publish Tx failed');
 
-	const packageId = ((publishTxn.objectChanges?.filter(
+	const id = ((publishTxn.objectChanges?.filter(
 		(a) => a.type === 'published',
 	)) ?? [])[0].packageId.replace(/^(0x)(0+)/, '0x');
```
milapsheth, line 56: fix this `replace` instead which is dropping the leading zero. Is that needed at all?

Foivos (approved):

scripts/publish-package.js
Foivos: it is not needed and it is now removed, it was originally part of the example publish scripts that Sui had online, because they want to delete the leading zeroes (making nicer looking package ids like `0xdee9` for deepbook)

Foivos (approved):

PR #25, opened 2024-05-02
fix(gateway): deserialize payload hash at the correct position

jcs47:
…correct position


milapsheth:
outdated due to gateway v2 update

PR #50, opened 2024-06-05
fix(axelar-gateway): check signer epoch for proof correctly

Foivos:


move/axelar_gateway/sources/auth.move
```diff
@@ -104,7 +104,7 @@ module axelar_gateway::auth {
         let current_epoch = self.epoch;
         let is_latest_signers = current_epoch == signers_epoch;
 
-        assert!(signers_epoch == 0 || (current_epoch - signers_epoch) >= PREVIOUS_KEY_RETENTION, EInvalidEpoch);
+        assert!(signers_epoch == 0 || (current_epoch - signers_epoch) <= PREVIOUS_KEY_RETENTION, EInvalidEpoch);
```
milapsheth, line 107: we need to take the `!` of the original expression
```suggestion
        assert!(signers_epoch != 0 && (current_epoch - signers_epoch) <= PREVIOUS_KEY_RETENTION, EInvalidEpoch);
```

milapsheth (approved):

PR #53, opened 2024-06-07
feat: cache objects

Foivos:


scripts/tx-builder.js
```diff
@@ -184,6 +199,7 @@ class TxBuilder {
                 ...options,
             },
         });
+        updateCache(result.objectChanges);
```
npty, line 202: If the cache is only used in the `getObject` function, can we update it directly there when there's a cache miss? btw, how does the performance differ between a cache hit and miss?

npty:
have you considered to turn this into a plugin? there's [a doc](https://sdk.mystenlabs.com/typescript/plugins#build-plugins) that i think it works pretty similar to this.

re1ro (approved):

PR #56, opened 2024-06-11
feat(operators): adding new operators contract

re1ro:
[AXE-4127]
* [x] `operators`: adding new operators contract for holding and borrowing capabilities

[AXE-4127]: https://axelarnetwork.atlassian.net/browse/AXE-4127?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


move/operators/sources/operators/operators.move
```diff
@@ -0,0 +1,212 @@
...
+
+        assert!(self.operator_ids.contains(&operator_id), 0);
+        assert!(self.caps.contains(cap_id), 1);
+
+        &self.caps[cap_id]
+    }
+
+    /// Removes a capability from the `Operators` struct.
+    public fun transfer_cap<T: key + store>(self: &mut Operators, _: &OwnerCap, cap_id: ID, new_holder: address) {
+        transfer::public_transfer(self.caps.remove<ID, T>(cap_id), new_holder);
```
Foivos, line 81: I suggest that this function just returns the cap instead and the owner can do as they please wtih it. Would probably rename this to get_cap or remove_cap as well.

move/operators/sources/operators/operators.move
```diff
@@ -0,0 +1,212 @@
...
+    }
+
+    /// Stores a capability in the `Operators` struct.
+    public fun store_cap<T: key + store>(self: &mut Operators, _: &OwnerCap, cap: T) {
+        let cap_id = object::id(&cap);
+        self.caps.add(cap_id, cap);
+    }
+
+    /// Allows an approved operator to borrow a capability by its ID.
+    public fun borrow_cap<T: key + store>(self: &Operators, operator_cap: &OperatorCap, cap_id: ID): &T {
```
Foivos, line 70: add a borrow_cap_mut probably too.

move/operators/sources/operators/operators.move
milapsheth, line 1: `operators` subfolder isn't needed. see other packages for reference

move/operators/sources/operators/operators.move
```diff
@@ -0,0 +1,319 @@
...
+
+        let removed_cap = remove_cap<OwnerCap>(&mut operators, &owner_cap, external_id);
+
+        destroy_operator_cap(operator_cap);
+        destroy_owner_cap(owner_cap);
+        destroy_owner_cap(external_cap);
+        destroy_owner_cap(removed_cap);
+        destroy_operators(operators);
+    }
+}
```
milapsheth, line 319: last line should have a newline, can you update your editor settings?

move/operators/sources/operators/operators.move
```diff
@@ -0,0 +1,319 @@
...
+    public struct OwnerCap has key, store {
+        id: UID,
+    }
+
+    /// The `OperatorCap` capability representing an approved operator.
+    public struct OperatorCap has key, store {
+        id: UID,
+    }
+
+    /// The main `Operators` struct storing the capabilities and operator IDs.
```
milapsheth, line 18: objects have a max size, so technically use of `vector` is bounded here. Add a comment that this is fine since number of operators in practice will be quite small and below Sui limits.

move/operators/sources/operators/operators.move
```diff
@@ -0,0 +1,319 @@
...
+    fun init(ctx: &mut TxContext) {
+        transfer::share_object(Operators {
+            id: object::new(ctx),
+            operator_ids: vector::empty(),
+            caps: bag::new(ctx),
+        });
+
+        transfer::public_transfer(OwnerCap {
+            id: object::new(ctx),
+        }, tx_context::sender(ctx));
```
milapsheth, line 39:
```suggestion
        let cap = OwnerCap {
            id: object::new(ctx),
        };

        transfer::transfer(cap, ctx.sender());
```
use move 2024 syntax. And refer to the pattern used in gateway contract, public_* isn't needed since object is defined in this module

move/operators/sources/operators/operators.move
```diff
@@ -0,0 +1,319 @@
...
+
+    // -----
+    // Setup
+    // -----
+
+    /// Initializes the contract and transfers the `OwnerCap` to the sender.
+    fun init(ctx: &mut TxContext) {
+        transfer::share_object(Operators {
+            id: object::new(ctx),
+            operator_ids: vector::empty(),
```
milapsheth, line 33:
```suggestion
            operator_ids: vector[],
```

move/operators/sources/operators/operators.move
```diff
@@ -0,0 +1,319 @@
...
+    /// The `OperatorCap` capability representing an approved operator.
+    public struct OperatorCap has key, store {
+        id: UID,
+    }
+
+    /// The main `Operators` struct storing the capabilities and operator IDs.
+    public struct Operators has key {
+        id: UID,
+        operator_ids: vector<ID>,
+        caps: Bag,
```
milapsheth, line 22: add comment of what this represents

move/operators/sources/operators/operators.move
```diff
@@ -0,0 +1,319 @@
...
+
+    /// The `OperatorCap` capability representing an approved operator.
+    public struct OperatorCap has key, store {
+        id: UID,
+    }
+
+    /// The main `Operators` struct storing the capabilities and operator IDs.
+    public struct Operators has key {
+        id: UID,
+        operator_ids: vector<ID>,
```
milapsheth, line 21:
```suggestion
        operators: VecSet<address>,
```
Use a `VecSet` and use `address` instead of ID since they're expected to be relayer addresses

move/operators/sources/operators/operators.move
```diff
@@ -0,0 +1,319 @@
...
+    // Public Functions
+    // ----------------
+
+    /// Adds a new operator by issuing an `OperatorCap` and storing its ID.
+    public fun add_operator(self: &mut Operators, _: &OwnerCap, new_operator: address, ctx: &mut TxContext) {
+        let operator_cap = OperatorCap {
+            id: object::new(ctx),
+        };
+        let operator_id = object::id(&operator_cap);
+        transfer::public_transfer(operator_cap, new_operator);
```
milapsheth, line 52:
```suggestion
        transfer::transfer(operator_cap, new_operator);
```
same comment

move/operators/sources/operators/operators.move
```diff
@@ -0,0 +1,319 @@
...
+        self.operator_ids.push_back(operator_id);
```
milapsheth, line 53: events are missing

move/operators/sources/operators/operators.move
```diff
@@ -0,0 +1,319 @@
...
+    }
+
+    /// Removes an operator by ID, revoking their `OperatorCap`.
+    public fun remove_operator(self: &mut Operators, _: &OwnerCap, operator_id: ID) {
+        let (found, index) = self.operator_ids.index_of(&operator_id);
+        assert!(found, 0);
```
milapsheth, line 59: create error constants

move/operators/sources/operators/operators.move
```diff
@@ -0,0 +1,319 @@
...
+        let operator_id = object::id(operator_cap);
+
+        assert!(self.operator_ids.contains(&operator_id), 0);
+        assert!(self.caps.contains(cap_id), 1);
+
+        &self.caps[cap_id]
+    }
+
+    /// Allows an approved operator to borrow a capability by its ID.
+    public fun borrow_cap_mut<T: key + store>(self: &mut Operators, operator_cap: &OperatorCap, cap_id: ID): &T {
```
milapsheth, line 80:
```suggestion
    public fun borrow_cap_mut<T: key + store>(self: &mut Operators, operator_cap: &OperatorCap, cap_id: ID): &mut T {
```

move/operators/sources/operators/operators.move
```diff
@@ -0,0 +1,319 @@
...
+        assert!(self.caps.contains(cap_id), 1);
+
+        &mut self.caps[cap_id]
+    }
+
+    /// Removes a capability from the `Operators` struct.
+    public fun remove_cap<T: key + store>(self: &mut Operators, _: &OwnerCap, cap_id: ID): T {
+        assert!(self.caps.contains(cap_id), 0);
+
+        self.caps.remove<ID, T>(cap_id)
```
milapsheth, line 93:
```suggestion
        self.caps.remove<ID, T>(cap_id)
```
remove will already abort if cap isn't present

move/operators/sources/operators/operators.move
re1ro: Removed.
I was looking mainly into gas_service. And it does have an extra folder, so does governance and squid.

move/operators/sources/operators/operators.move
re1ro: Added the new line. My editor is not enforcing anything there 😃

move/operators/sources/operators/operators.move
milapsheth: i see, we can remove for those

move/operators/sources/operators/operators.move
milapsheth: yeah, i meant to make your editor add a trailing new line on save or something

move/operators/sources/operators.move
```diff
@@ -18,10 +20,46 @@ module operators::operators {
     /// The main `Operators` struct storing the capabilities and operator IDs.
     public struct Operators has key {
         id: UID,
-        operator_ids: vector<ID>,
+        // number of operators should small and under the Sui vector limits
+        operators: VecSet<ID>,
```
milapsheth, line 24:
```suggestion
        operators: VecSet<address>,
```
missing change to address instead of ID

move/operators/sources/operators.move
```diff
@@ -18,10 +20,46 @@ module operators::operators {
     /// The main `Operators` struct storing the capabilities and operator IDs.
     public struct Operators has key {
         id: UID,
-        operator_ids: vector<ID>,
+        // number of operators should small and under the Sui vector limits
```
milapsheth, line 23:
```suggestion
        // The number of operators are small in practice, and under the Sui object size limit, so a dynamic collection doesn't need to be used
```
Let's try to be more precise so it makes sense to someone reading this a year later

move/operators/sources/operators.move
```diff
@@ -18,10 +20,46 @@ module operators::operators {
...
+        operators: VecSet<ID>,
+        // map-like collection of capabilities stored as Sui objects
         caps: Bag,
     }
 
+    // -----
+    // Error constants
+    // -----
```
milapsheth, line 31:
```suggestion
    // ------
    // Errors
    // ------
```

move/operators/sources/operators.move
```diff
@@ -18,10 +20,46 @@ module operators::operators {
...
+
+    /// When the operator is not found in the set of approved operators.
+    const EOperatorNotFound: u64 = 0;
+
+    /// When the capability is not found.
+    const ECapNotFound: u64 = 1;
+
+    // -----
+    // Events
+    // -----
```
milapsheth, line 41:
```suggestion
    // ------
    // Events
    // ------
```

move/operators/sources/operators.move
```diff
@@ -18,10 +20,46 @@ module operators::operators {
...
+
+    /// Event emitted when a new operator is added.
+    public struct OperatorAdded has copy, drop {
+        operator_id: ID,
```
milapsheth, line 45: use address and not ids

move/operators/sources/operators.move
```diff
@@ -18,10 +20,46 @@ module operators::operators {
...
+    }
+
+    /// Event emitted when an operator is removed.
+    public struct OperatorRemoved has copy, drop {
+        operator_id: ID,
+    }
+
+    /// Event emitted when a capability is stored.
+    public struct CapabilityStored has copy, drop {
+        cap_id: ID,
```
milapsheth, line 55: add type name in this as well

move/operators/sources/operators.move
```diff
@@ -101,22 +153,24 @@ module operators::operators {
...
 
-        while (!operator_ids.is_empty()) {
-            operator_ids.pop_back();
+        let mut keys = operators.into_keys();
+
+        while (!keys.is_empty()) {
+            keys.pop_back();
         };
-        operator_ids.destroy_empty();
+        keys.destroy_empty();
```
milapsheth, line 173:
```suggestion

        keys.destroy_empty();
```

move/operators/sources/operators.move
```diff
@@ -166,11 +220,11 @@ module operators::operators {
 
         let new_operator = @0x1;
         add_operator(&mut operators, &owner_cap, new_operator, ctx);
-        assert!(operators.operator_ids.length() == 1, 0);
+        assert!(operators.operators.keys().length() == 1, 0);
```
milapsheth, line 223:
```suggestion
        assert!(operators.operators.size() == 1, 0);
```

move/operators/sources/operators.move
re1ro: I though we agreed on the call to use Capabilities instead of the relaying on the address of TX origin. If we store the operator address instead how are we going to validate the Capability? Because removing the address from the list still allows to transfer the capability to another address and use it.

move/operators/sources/operators.move
re1ro: See above. I still think we should use IDs

move/operators/sources/operators.move
re1ro: But that address won't likely to be whitelisted. Ok I will make the change

move/operators/sources/operators.move
```diff
@@ -113,24 +117,31 @@ module operators::operators {
...
     }
 
     /// Allows an approved operator to borrow a capability by its ID.
-    public fun borrow_cap<T: key + store>(self: &Operators, operator_cap: &OperatorCap, cap_id: ID): &T {
-        let operator_id = object::id(operator_cap);
-
-        assert!(self.operators.contains(&operator_id), EOperatorNotFound);
+    public fun borrow_cap<T: key + store>(
+        self: &Operators,
+        _operator_cap: &OperatorCap,
```
milapsheth, line 127: why keep this unused arg? you can use also index syntax by removing this perhaps

milapsheth (approved):

move/operators/sources/operators.move
re1ro: This is what checks access through capability verification, just check on the `sender` is not enough. Because it's like `tx.origin` and not like `msg.sender`.

PR #60, opened 2024-06-18
feat: convert all scripts to ts and removed unused scripts

Foivos:


npty:
Could you please add TS linting tool as well? I recommend starting with https://typescript-eslint.io/.

src/tx-builder.ts
```diff
@@ -0,0 +1,314 @@
...
+            (
+                change as {
+                    objectId: string;
+                }
+            ).objectId
+        ] = change;
+    }
+}
+
+function getObject(tx: TransactionBlock, object: TransactionObjectInput) {
```
npty, line 35: could you please explicitly declare the return type here an other functions as well?

src/tx-builder.ts
```diff
@@ -0,0 +1,314 @@
...
+
+        if (type.typeArguments.length > 0) {
+            name += `<${type.typeArguments[0]}`;
+
+            for (let i = 1; i < type.typeArguments.length; i++) {
+                name += `,${type.typeArguments[i]}`;
+            }
+
+            name += '>';
+        }
```
npty, line 69: we can utilize [array.join](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/join) here

src/tx-builder.ts
```diff
@@ -0,0 +1,314 @@
...
+
+    if ((type as { Struct: Type }).Struct) {
+        return get((type as { Struct: Type }).Struct);
+    } else if ((type as { Reference: SuiMoveNormalizedType }).Reference) {
+        return getTypeName((type as { Reference: SuiMoveNormalizedType }).Reference);
+    } else if ((type as { MutableReference: SuiMoveNormalizedType }).MutableReference) {
+        return getTypeName((type as { MutableReference: SuiMoveNormalizedType }).MutableReference);
+    } else if ((type as { Vector: SuiMoveNormalizedType }).Vector) {
+        return `vector<${getTypeName((type as { Vector: SuiMoveNormalizedType }).Vector)}>`;
+    }
```
npty, line 82: can you pls try using [this strategy](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#the-in-operator-narrowing) for narrowing the type? other strategies in the link can be useful in other places as well since sui has many union types in their sdk.

milapsheth (approved):

src/tx-builder.ts
```diff
@@ -0,0 +1,314 @@
...
+function updateCache(objectChanges: SuiObjectChange[]) {
+    for (const change of objectChanges) {
+        if (
+            !(
+                change as {
+                    objectId: string;
+                }
+            ).objectId
+        )
+            continue;
```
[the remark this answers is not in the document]
Foivos: oooh nice

PR #65, opened 2024-07-02
refactor: add option to generate unsigned tx for publishing package

blockchainguyy:
https://axelarnetwork.atlassian.net/browse/AXE-4129

Needed a new function to return module dependencies and bytecode.

### DIFF
--- a/scripts/publish-package.js
+++ b/scripts/publish-package.js
@@ -19,7 +19,7 @@ async function getContractBuild(packageName) {
 
     const tmpobj = tmp.dirSync({ unsafeCleanup: true });
 
-    return JSON.parse(
+    const base64ByteCode = JSON.parse(
         execSync(
             `sui move build --dump-bytecode-as-base64 --path ${path.join(__dirname, '/../move/', packageName)} --install-dir ${
                 tmpobj.name
@@ -30,6 +30,8 @@ async function getContractBuild(packageName) {
             },
         ),
     );
+
+    return base64ByteCode;
 }
 
 async function publishPackage(packageName, client, keypair, options = {}) {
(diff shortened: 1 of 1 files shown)


re1ro (approved):

scripts/publish-package.js
```diff
@@ -38,25 +42,39 @@ async function publishPackage(packageName, client, keypair) {
     });
 
     // Transfer the upgrade capability to the sender so they can upgrade the package later if they want.
+    let address;
+
+    if (!options.sender) {
+        address = keypair.getPublicKey().toSuiAddress();
+    } else {
+        address = options.sender;
+    }
```
milapsheth, line 51:
```suggestion
    const address = options.sender || keypair.getPublicKey().toSuiAddress();
```
simplify

scripts/publish-package.js
```diff
@@ -38,25 +42,39 @@ async function publishPackage(packageName, client, keypair) {
...
-    console.info(`Published package ${packageId} from address ${address}}`);
+        console.info(`Published package ${packageId} from address ${address}}`);
 
-    return { packageId, publishTxn };
+        return { packageId, publishTxn };
+    }
+
+    tx.setSenderIfNotSet(address);
+    const txBytes = await tx.build({ client });
+    return toB64(txBytes);
```
milapsheth, line 77: return types are different based on inputs, generally a bad practice. return an object with another field

scripts/publish-package.js
```diff
@@ -38,25 +42,39 @@ async function publishPackage(packageName, client, keypair) {
...
-        signer: keypair,
-        options: {
-            showEffects: true,
-            showObjectChanges: true,
-            showContent: true,
-        },
-        requestType: 'WaitForLocalExecution',
-    });
-    if (publishTxn.effects?.status.status !== 'success') throw new Error('Publish Tx failed');
+    if (!options.offline) {
```
milapsheth, line 55: instead of nesting the complicated block of code, do the easier part first, in this case, the offline case

milapsheth, line 55: what's the use case for offline support for publishing packages? similar to evm, i don't see why we're offline signing deployments as opposed to other txs. I'm ok to add it since it's a small change and it's already ready, but otherwise best to avoid spending time on use cases we don't need

scripts/publish-package.js
```diff
@@ -3,23 +3,23 @@ const { setConfig, getFullObject, requestSuiFromFaucet } = require('./utils');
...
     fs.writeFileSync(`${__dirname}/../move/${packageName}/Move.toml`, fillAddresses(toml, '0x0', packageName));
 
     // remove all controlled temporary objects on process exit
-    const address = keypair.getPublicKey().toSuiAddress();
     tmp.setGracefulCleanup();
 
     const tmpobj = tmp.dirSync({ unsafeCleanup: true });
 
-    const { modules, dependencies } = JSON.parse(
+    return JSON.parse(
```
milapsheth, line 22: avoid returning a complex expression, instead assign it to a variable so it makes more sense, and return that

milapsheth (approved):

milapsheth:
functionality has been moved into TS scripts, this needs some refactor

PR #74, opened 2024-07-17
feat: add uid type

npty:
added uid type

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -1,0 +1,7 @@
+const { bcs } = require('@mysten/sui.js/bcs');
+const { fromHEX, toHEX } = require('@mysten/bcs');
+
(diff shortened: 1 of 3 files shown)


src/types.ts
```diff
@@ -0,0 +1,7 @@
+const { bcs } = require('@mysten/sui.js/bcs');
+const { fromHEX, toHEX } = require('@mysten/bcs');
+
+export const UID = bcs.fixedArray(32, bcs.u8()).transform({
```
milapsheth, line 4: could aliasing this with bcs.Address be enough instead of redefining it?

milapsheth (approved):

re1ro (approved):

milapsheth (approved):

PR #85, opened 2024-07-23
chore: add release snapshot workflow

npty:
# Description

This PR add release snapshot workflow to release unofficial version when you need it for development. The published version format will be `@axelar-network/axelar-cgp-sui@0.0.0-<commit-sha>`


.github/workflows/release-snapshot.yaml
```diff
@@ -0,0 +1,34 @@
...
+      - name: Build TS and SUI modules
+        run: npm run build
+
+      - name: Set Snapshot Version
+        run: |
+          npm version 0.0.0-snapshot.$(git rev-parse --short HEAD) --git-tag-version=false
+
+      - name: Publish Snapshot to NPM
+        run: |
+          echo "//registry.npmjs.org/:_authToken=${{ secrets.NPM_TOKEN }}" > .npmrc
```
milapsheth, line 33: why not just set the registry during the node setup?

milapsheth (approved):

.github/workflows/release-snapshot.yaml
npty: have a better approach here https://github.com/axelarnetwork/axelar-cgp-sui/pull/85/commits/ddd1723c6f1bbc22a25c59078c8e27549b2f6c43

PR #87, opened 2024-07-31
feat: its hub support

Foivos:
[AXE-4584]

[AXE-4584]: https://axelarnetwork.atlassian.net/browse/AXE-4584?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


move/its/sources/service.move
```diff
@@ -273,20 +286,37 @@ module its::service {
...
+            mut source_chain,
             _,
             source_address,
-            payload
+            mut payload
         ) = self.channel_mut().consume_approved_message(approved_message);
 
         assert!(self.is_trusted_address(source_chain, source_address), EUntrustedAddress);
 
+        if(source_chain.into_bytes() == AXELAR_CHAIN_NAME) {
```
milapsheth, line 297:
```suggestion
        if (source_chain.into_bytes() == AXELAR_CHAIN_NAME) {
```
and in other places

move/its/sources/service.move
```diff
@@ -273,20 +286,37 @@ module its::service {
...
+            let mut reader = abi::new_reader(payload);
+            if(reader.read_u256() == MESSAGE_TYPE_RECEIVE_FROM_HUB) {
```
milapsheth, line 299: logic doesn't match EVM: https://github.com/axelarnetwork/interchain-token-service/pull/265/files#diff-0617e1ea35b71e4e1254741ae6248a79b107ff70dacc130cdedbbf3247daeb51R923

check if message type is receive from hub first. Then assert that source_chain is Axelarnet

milapsheth: I think we need to change the EVM reference slightly. If you're sending/receiving ITS transfer to/from Axelarnet itself, I think we should still wrap/unwrap via the ITS hub message type to make the hub payloads easier to reason about. Let's change the logic to, if the trusted address is set to `"hub"`, then set destination chain to `Axelarnet`, and get its-hub address via `trustedAddress["its-hub"]` so `Axelarnet` trusted address can be set to `"hub"`

move/its/sources/service.move
```diff
@@ -27,6 +27,19 @@ module its::service {
...
+     */
+    const AXELAR_CHAIN_NAME: vector<u8> = b"Axelarnet";
+
+    /**
+     * @dev Special trusted address value that indicates that the ITS call
+     * for that destination chain should be routed via the ITS hub.
+     */
+    const ITS_HUB_TRUSTED_ADDRESS: vector<u8> = b"hub";
 
     // address::to_u256(address::from_bytes(keccak256(b"sui-set-trusted-addresses")));
```
milapsheth, line 44: can you add some tests as well please?

move/its/sources/service.move
```diff
@@ -273,20 +287,45 @@ module its::service {
...
+
+        // Check whether the ITS call should be routed via ITS hub for this destination chain
+        if(destination_address.into_bytes() == ITS_HUB_ROUTING_IDENTIFIER) {
+            let mut writter = abi::new_writer(3);
+            writter.write_u256(MESSAGE_TYPE_SEND_TO_HUB);
+            writter.write_bytes(destination_chain.into_bytes());
+            writter.write_bytes(payload);
+            payload = writter.into_bytes();
+            destination_chain = ascii::string(ITS_HUB_CHAIN_NAME);
+            destination_address = self.get_trusted_address(destination_chain);
```
npty, line 323: Doesn't this produce the same result as the initial assignment at the top of the function?

move/its/sources/service.move
```diff
@@ -273,20 +287,45 @@ module its::service {
...
+        };
+
+        // Check whether no trusted address was set for the destination chain
+        assert!(destination_address.length() > 0, EUntrustedChain);
```
npty, line 327: perhaps move this assertion to the top of the function, after the initial assignment? (if the value is the same as re-assignment value)

move/its/sources/service.move
```diff
@@ -273,20 +287,45 @@ module its::service {
...
+
         gateway::call_contract(self.channel_mut(), destination_chain, destination_address, payload);
```
npty, line 329: perhaps call the gas service contract to pay gas here as well?

move/its/sources/service.move
```diff
@@ -27,6 +27,18 @@ module its::service {
     const MESSAGE_TYPE_INTERCHAIN_TRANSFER: u256 = 0;
     const MESSAGE_TYPE_DEPLOY_INTERCHAIN_TOKEN: u256 = 1;
     //const MESSAGE_TYPE_DEPLOY_TOKEN_MANAGER: u256 = 2;
+    const MESSAGE_TYPE_SEND_TO_HUB: u256 = 3;
+    const MESSAGE_TYPE_RECEIVE_FROM_HUB: u256 = 4;
+
+    /**
+     * @dev Chain name for Axelar. This is used for routing ITS calls via ITS hub on Axelar.
+     */
```
milapsheth, line 35:
```suggestion
    // Chain name for Axelar. This is used for routing ITS calls via ITS hub on Axelar.
```
not natspec

move/its/sources/service.move
```diff
@@ -27,6 +27,18 @@ module its::service {
...
+    const ITS_HUB_CHAIN_NAME: vector<u8> = b"Axelarnet";
+
+    /**
+     * @dev Identifier to be used as destination address for chains that route to hub. For Sui this will probably be every supported chain.
```
milapsheth, line 39: same

move/its/sources/service.move
```diff
@@ -273,20 +287,45 @@ module its::service {
...
-            payload
+            mut payload
         ) = self.channel_mut().consume_approved_message(approved_message);
 
         assert!(self.is_trusted_address(source_chain, source_address), EUntrustedAddress);
 
+        let mut reader = abi::new_reader(payload);
+        if (reader.read_u256() == MESSAGE_TYPE_RECEIVE_FROM_HUB) {
+            assert!(source_chain.into_bytes() == ITS_HUB_CHAIN_NAME, ESenderNotHub);
+            source_chain = ascii::string(reader.read_bytes());
```
milapsheth, line 301:
```suggestion
            assert!(source_chain.into_bytes() == ITS_HUB_CHAIN_NAME, ESenderNotHub);

            source_chain = ascii::string(reader.read_bytes());
```

milapsheth, line 301: Add a `read_string` to Abi package that returns a ascii::string for clarity

milapsheth, line 301: Missing check for whether original source chain is registered correctly
https://github.com/axelarnetwork/interchain-token-service/blob/main/contracts/InterchainTokenService.sol#L978-L981

move/its/sources/service.move
```diff
@@ -273,20 +287,45 @@ module its::service {
...
-    fun send_payload(self: &mut ITS, destination_chain: String, payload: vector<u8>) {
-        let destination_address = self.get_trusted_address(destination_chain);
+    fun send_payload(self: &mut ITS, mut destination_chain: String, mut payload: vector<u8>) {
+        let mut destination_address = self.get_trusted_address(destination_chain);
+
+        // Prevent sending directly to the ITS Hub chain. This is not supported yet, so fail early to prevent the user from having their funds stuck.
+        assert!(destination_chain.into_bytes() != ITS_HUB_CHAIN_NAME, EUntrustedChain);
+
+        // Check whether the ITS call should be routed via ITS hub for this destination chain
+        if(destination_address.into_bytes() == ITS_HUB_ROUTING_IDENTIFIER) {
```
milapsheth, line 316:
```suggestion
        if (destination_address.into_bytes() == ITS_HUB_ROUTING_IDENTIFIER) {
```

move/its/sources/service.move
```diff
@@ -273,20 +287,45 @@ module its::service {
...
+            let mut writter = abi::new_writer(3);
+            writter.write_u256(MESSAGE_TYPE_SEND_TO_HUB);
+            writter.write_bytes(destination_chain.into_bytes());
```
milapsheth, line 319:
```suggestion
            writter.write_string(destination_chain);
```
add ascii string write support

move/its/sources/service.move
```diff
@@ -273,20 +287,45 @@ module its::service {
...
-        let destination_address = self.get_trusted_address(destination_chain);
+    fun send_payload(self: &mut ITS, mut destination_chain: String, mut payload: vector<u8>) {
+        let mut destination_address = self.get_trusted_address(destination_chain);
+
+        // Prevent sending directly to the ITS Hub chain. This is not supported yet, so fail early to prevent the user from having their funds stuck.
+        assert!(destination_chain.into_bytes() != ITS_HUB_CHAIN_NAME, EUntrustedChain);
+
+        // Check whether the ITS call should be routed via ITS hub for this destination chain
+        if(destination_address.into_bytes() == ITS_HUB_ROUTING_IDENTIFIER) {
+            let mut writter = abi::new_writer(3);
```
milapsheth, line 317:
```suggestion
            let mut writer = abi::new_writer(3);
```

move/its/sources/service.move
milapsheth, line 327: this should never occur since we don't allow storing empty strings, and we explicitly check if the value exists in the table

move/its/sources/service.move
milapsheth: there's another PR with this functionality

move/its/sources/discovery.move
```diff
@@ -66,9 +68,16 @@ module its::discovery {
...
+
+        if (message_type == MESSAGE_TYPE_RECEIVE_FROM_HUB) {
+            reader.skip_slot();
+            payload = reader.read_bytes();
+            reader = abi::new_reader(payload);
+            message_type = reader.read_u256();
+        };
 
         if (message_type == MESSAGE_TYPE_INTERCHAIN_TRANSFER) {
             get_interchain_transfer_tx(self, &mut reader)
```
milapsheth, line 83:
```suggestion
            return get_interchain_transfer_tx(self, &mut reader)
```
return early, so else nesting isn't needed

milapsheth (approved):

move/its/sources/service.move
Foivos: We could do the same for `u64`, `bool`, `vector<String>` and countless other types, where do we draw the line?

move/its/sources/service.move
Foivos: same as above, where do we stop?

move/its/sources/service.move
Foivos: fixed

move/its/sources/service.move
Foivos: removed

move/its/sources/service.move
Foivos: fixed

move/its/sources/discovery.move
Foivos: Can you please explain? I thought in rust if you had an if statement at the end of a function without a semicolon, whatever the first successful branch returns is returned as well, meaning changing to what you suggest makes zero difference. Where am I wrong?

move/its/sources/service.move
```diff
@@ -30,14 +30,10 @@ module its::service {
...
-    /**
-     * @dev Chain name for Axelar. This is used for routing ITS calls via ITS hub on Axelar.
-     */
+    // Chain name for Axelar. This is used for routing ITS calls via ITS hub on Axelar.
     const ITS_HUB_CHAIN_NAME: vector<u8> = b"Axelarnet";
 
-    /**
-     * @dev Identifier to be used as destination address for chains that route to hub. For Sui this will probably be every supported chain.
-     */
+    //Identifier to be used as destination address for chains that route to hub. For Sui this will probably be every supported chain.
```
milapsheth, line 36:
```suggestion
    // Identifier to be used as destination address for chains that route to hub. For Sui this will probably be every supported chain.
```

move/its/sources/service.move
```diff
@@ -297,9 +292,14 @@ module its::service {
 
         let mut reader = abi::new_reader(payload);
         if (reader.read_u256() == MESSAGE_TYPE_RECEIVE_FROM_HUB) {
-            assert!(source_chain.into_bytes() == ITS_HUB_CHAIN_NAME, ESenderNotHub);
+            assert!(source_chain.into_bytes() == ITS_HUB_CHAIN_NAME, EUntrustedChain);
+
             source_chain = ascii::string(reader.read_bytes());
             payload = reader.read_bytes();
+
+            assert!(self.trusted_addresses(source_chain) == ITS_HUB_ROUTING_IDENTIFIER, EUntrustedChain);
```
milapsheth, line 300:
```suggestion
            assert!(self.trusted_address(source_chain) == ITS_HUB_ROUTING_IDENTIFIER, EUntrustedChain);
```

scripts/upgradable.js
milapsheth, line 1: this was checked in incorrectly

move/its/sources/service.move
milapsheth: We should add it for any type that we use. If it makes it cleaner, then why not. Internally, `read_string` can use `read_bytes`, but the caller should avoid dealing with the complexity.

move/its/sources/discovery.move
Foivos: You cannot have the first one, since the last branch does not return the right type

PR #89, opened 2024-08-03
feat: its tests actual

Foivos:


move/its/sources/its.move
```diff
@@ -77,6 +78,19 @@ module its::its {
         self.address_tracker.set_trusted_address(chain_name, trusted_address);
     }
 
+    public(package) fun set_trusted_addresses(self: &mut ITS, trusted_addresses: TrustedAddresses) {
+        let (mut chain_names, mut trusted_addresses) = trusted_addresses.destroy();
+        let length = chain_names.length();
+        let mut i = 0;
+        while(i < length) {
+            self.set_trusted_address(
+                ascii::string(chain_names.pop_back()),
```
milapsheth, line 87: Use string types in TrustedAddresses to avoid the separate cast?

move/its/sources/trusted_addresses.move
```diff
@@ -0,0 +1,37 @@
+module its::trusted_addresses {
+    use sui::bcs::BCS;
+
+    const EMalformedTrustedAddresses: u64 = 0;
+
+    public struct TrustedAddresses has copy, drop {
+        trusted_chains: vector<vector<u8>>,
```
milapsheth, line 7: Use vector<string>

move/its/tests/coin.move
```diff
@@ -0,0 +1,48 @@
+#[test_only]
+module its::coin {
+    use sui::url::{Url};
+    use sui::coin::{Self, CoinMetadata, TreasuryCap};
+
+    public struct COIN has drop{}
+    
+
+    public fun create_treasury_and_metadata(symbol: vector<u8>, decimals: u8, ctx: &mut TxContext): (TreasuryCap<COIN>, CoinMetadata<COIN>) {
```
milapsheth, line 9: Add section comments (similar to other modules to highlight when sections start)

move/its/sources/trusted_addresses.move
```diff
@@ -0,0 +1,37 @@
...
+    }
+
+    #[test_only]
+    public fun new_for_testing(trusted_chains: vector<vector<u8>>, trusted_addresses: vector<vector<u8>>): TrustedAddresses {
+        TrustedAddresses {
+            trusted_chains,
+            trusted_addresses,
+        }
+    }
+}
```
milapsheth, line 37: Missing newline at the end (configure your editor to add trailing newline by default?)

move/its/sources/trusted_addresses.move
```diff
@@ -0,0 +1,37 @@
...
+            trusted_addresses,
+        }
+    }
+
+    public fun destroy(self: TrustedAddresses): (vector<vector<u8>>, vector<vector<u8>>) {
+        let TrustedAddresses { trusted_chains, trusted_addresses } = self;
+        (trusted_chains, trusted_addresses)
+    }
+
+    #[test_only]
```
milapsheth, line 30: Add section comment to separate when test method starts

milapsheth (approved):

move/governance/sources/governance/governance.move
```diff
@@ -123,37 +123,49 @@ use sui::test_scenario;
...
+    ctx: &mut TxContext,
+): Governance {
+    Governance {
+        id: object::new(ctx),
+        trusted_source_chain,
+        trusted_source_address,
+        message_type,
+        channel: channel::new(ctx),
+        caps: table::new<ID, UpgradeCap>(ctx),
+    }
```
milapsheth, line 140: Add its hub test coverage?

move/its/sources/its.move
Foivos: Because we are peeling of off bcs, we can `peel_vec_vec_u8` but we can't `peel_vec_string` which is why I chose vectors which needs the cast.

move/its/sources/trusted_addresses.move
Foivos: see above comment.

move/governance/sources/governance/governance.move
Foivos: See lines 1329-1435

move/its/sources/trusted_addresses.move
Foivos: done

move/its/sources/trusted_addresses.move
Foivos: done

move/its/tests/coin.move
Foivos: done

move/its/sources/coin_management.move
```diff
@@ -26,6 +29,9 @@ module its::coin_management {
         dust: u256,
     }
 
+    // ------
+    // Functions to create CoinManagement
```
milapsheth, line 33:
```suggestion
    // Public Functions to create CoinManagement
```
to clarify from internal functions section

move/its/sources/utils.move
```diff
@@ -83,6 +83,7 @@ module its::utils {
         res
     }
     
+    // === Tests ===
```
milapsheth, line 86:
```suggestion
    // -----
    // Tests
    // -----
```
for consistency

move/its/tests/coin.move
```diff
@@ -0,0 +1,52 @@
+#[test_only]
+module its::coin {
+    use sui::url::{Url};
+    use sui::coin::{Self, CoinMetadata, TreasuryCap};
+
+    public struct COIN has drop{}
```
milapsheth, line 6:
```suggestion
    public struct COIN has drop {}
```

move/its/sources/its.move
milapsheth: I'm suggesting to add `peel_vec_string` in abi so it's cleaner to work with here

move/its/sources/trusted_addresses.move
```diff
@@ -0,0 +1,38 @@
...
+    use sui::bcs::BCS;
+
+    const EMalformedTrustedAddresses: u64 = 0;
+
+    public struct TrustedAddresses has copy, drop {
+        trusted_chains: vector<vector<u8>>,
+        trusted_addresses: vector<vector<u8>>,
+    }
+
+    public fun peel(bcs: &mut BCS): TrustedAddresses {
```
milapsheth, line 11: Instead of adding a special `TrustedAddresses` struct, can we just submit multiple set_trusted_address calls? Doesn't need to block this PR though

milapsheth (approved):

move/its/sources/trusted_addresses.move
Foivos: since these are each a gateway Message, it would be a bit of a pain for axelar to manage, and would make it more prone to errors.

Foivos: Also, for the abi comment, it is BCS that does not have peel_vec_string, not abi.

move/its/sources/its.move
Foivos: But we are peeling of off bcs, not abi. Are you suggesting changing the data input entirely?

milapsheth: Ah, missed that part. I'm ok with keeping it as BCS, but we should still clean this up. Let's get to it after the call contract ticket refactor

PR #96, opened 2024-08-09
feat: use hot potato pattern to borrow cap

npty:
# Description

[AXE-4628](https://axelarnetwork.atlassian.net/browse/AXE-4628)

This PR fix issue on the `borrow_cap` function, in the `operators` contract. Previously, it returns a reference to the `cap<T>` object which is not working on the client side using the PTBs. 

Instead of returning a reference to cap object, we return a hot potato object `BorrowedCap<T>` which is introduced in this PR. Since it's a hot potato object which doesn't have any abilities, we also need to add a function `restore_cap` to allow consuming it within the same transaction.

[AXE-4628]: https://axelarnetwork.atlassian.net/browse/AXE-4628?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/move/operators/sources/operators.move
+++ b/move/operators/sources/operators.move
@@ -142,7 +142,7 @@ module operators::operators {
         // Create a new `Referent` to store the loaned capability
         let mut referent = borrow::new(cap, ctx);
 
-        // Borrow the T capability and a Borrow hot potato object from the `Referent`
+        // Create a `Borrow` hot potato object from the `Referent` that needs to be returned within the same tx
         let (loaned_cap, borrow_obj) = borrow::borrow(&mut referent);
 
         // Store the `Referent` in the `Operators` struct
--- a/move/operators/sources/operators.move
+++ b/move/operators/sources/operators.move
@@ -28,7 +28,7 @@ module operators::operators {
         caps: Bag,
     }
 
-    public struct BorrowedCap<T> {
+    public struct BorrowedCap<T: key + store> {
         cap: T
     }
 
@@ -158,11 +158,16 @@ module operators::operators {
         _operator_cap: &OperatorCap,
         cap_id: ID,
         ctx: &mut TxContext
-    ): &mut T {
+    ): BorrowedCap<T> {
         assert!(self.operators.contains(&ctx.sender()), EOperatorNotFound);
         assert!(self.caps.contains(cap_id), ECapNotFound);
 
-        &mut self.caps[cap_id]
+        // &mut self.caps[cap_id]
+        let caps = bag::remove(&mut self.caps, cap_id);
+
+        BorrowedCap<T> {
+          cap: caps
+        }
     }
 
     /// Removes a capability from the `Operators` struct.
--- a/move/operators/sources/operators.move
+++ b/move/operators/sources/operators.move
@@ -122,7 +122,8 @@ module operators::operators {
         });
     }
 
-    /// Allows an approved operator to loan out a capability by its ID. The loaned capability needs to be restored by the end of the transaction.
+    /// Allows an approved operator to temporarily loan out a capability by its ID.
+    /// The loaned capability must be restored by the end of the transaction.
     public fun loan_cap<T: key + store>(
         self: &mut Operators,
         _operator_cap: &OperatorCap,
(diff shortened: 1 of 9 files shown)


move/operators/sources/operators.move
```diff
@@ -28,6 +28,10 @@ module operators::operators {
         caps: Bag,
     }
 
+    public struct BorrowedCap<T: key + store> {
```
milapsheth, line 31:
```suggestion
    public struct LoanedCap<T: key + store> {
```
maybe?

move/operators/sources/operators.move
```diff
@@ -123,15 +127,29 @@ module operators::operators {
 
     /// Allows an approved operator to borrow a capability by its ID.
     public fun borrow_cap<T: key + store>(
```
milapsheth, line 129:
```suggestion
    /// Allows an approved operator to loan out a capability by its ID. The loaned capability needs to be restored by the end of the transaction.
    public fun loan_cap<T: key + store>(
```

move/operators/sources/operators.move
```diff
@@ -123,15 +127,29 @@ module operators::operators {
...
         assert!(self.operators.contains(&ctx.sender()), EOperatorNotFound);
         assert!(self.caps.contains(cap_id), ECapNotFound);
 
-        &self.caps[cap_id]
+        BorrowedCap<T> {
+            cap: bag::remove(&mut self.caps, cap_id)
+        }
+    }
+
+    /// Restore a borrowed capability back to the `Operators` struct.
```
milapsheth, line 143:
```suggestion
    /// Restore a loaned capability back to the `Operators` struct.
```

move/operators/sources/operators.move
```diff
@@ -121,30 +122,55 @@ module operators::operators {
...
 
-        &self.caps[cap_id]
+        // Remove the capability from the `Operators` struct to loan it out
+        let cap = self.caps.remove<ID, T>(cap_id);
+
+        // Create a new `Referent` to store the loaned capability
+        let mut referent = borrow::new<T>(cap, ctx);
+
+        // Borrow the T capability and a Borrow hot potato object from the `Referent`
+        let (loaned_cap, borrow_obj) = borrow::borrow<T>(&mut referent);
```
milapsheth, line 143:
```suggestion
        let (loaned_cap, borrow_obj) = borrow::borrow(&mut referent);
```
Is the T qualifier needed or is it inferred automatically?

move/operators/sources/operators.move
```diff
@@ -121,30 +122,55 @@ module operators::operators {
...
+
+        // Store the `Referent` back in the `Operators` struct
+        self.caps.add(cap_id, referent);
```
milapsheth, line 146: Store the referent in a separate `self.loaned_caps` bag. Otherwise the operator can loan out the same cap_id again in this tx wrapping the Referent within another Referent. Technically, it can still work since the borrows have to be restored in the same sequence for it to work, but I don't like allowing this weird behaviour.

move/operators/sources/operators.move
```diff
@@ -121,30 +122,55 @@ module operators::operators {
...
         assert!(self.caps.contains(cap_id), ECapNotFound);
 
-        &self.caps[cap_id]
+        // Remove the capability from the `Operators` struct to loan it out
+        let cap = self.caps.remove<ID, T>(cap_id);
+
+        // Create a new `Referent` to store the loaned capability
+        let mut referent = borrow::new<T>(cap, ctx);
+
+        // Borrow the T capability and a Borrow hot potato object from the `Referent`
```
milapsheth, line 142:
```suggestion
        // Create a `Borrow` hot potato object from the `Referent` that needs to be returned within the same tx
```

move/operators/sources/operators.move
npty: automatically

move/operators/sources/operators.move
```diff
@@ -26,6 +27,8 @@ module operators::operators {
         operators: VecSet<address>,
         // map-like collection of capabilities stored as Sui objects
         caps: Bag,
+        // map-like collection of Referents storing loaned capabilities
```
milapsheth, line 30:
```suggestion
        // map-like collection of Referents storing loaned capabilities. Referents only get stored for the duration of the tx
```

milapsheth (approved):

PR #109, opened 2024-08-23
chore: format by Prettier Move plugin

npty:
# Description

This PR manually formats all move files by using "Prettier Move" plugin. More details in [AXE-4683](https://axelarnetwork.atlassian.net/browse/AXE-4683)

[AXE-4683]: https://axelarnetwork.atlassian.net/browse/AXE-4683?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Todo
- [x] Make it compilable


.github/workflows/test-js.yaml
```diff
@@ -12,7 +12,7 @@ concurrency:
   cancel-in-progress: true
 
 env:
-  SUI_VERSION: mainnet-v1.25.3
+  SUI_VERSION: mainnet-v1.31.1
```
milapsheth, line 15: can we read this from a single place to avoid duplication? Might be useful to add a warning in hardhat.config.js as well if a different sui version is being used locally

npty: good idea, but definitely not in the hardhat file 😅

milapsheth (approved):

move/abi/Move.lock
```diff
@@ -21,6 +21,6 @@ dependencies = [
 ]
 
 [move.toolchain-version]
-compiler-version = "1.26.1"
+compiler-version = "1.31.1"
```
milapsheth, line 24: Let's use testnet-1.32 version. It has enum support

PR #115, opened 2024-08-28
Integrate DeepBook v3

0xaslan:


move/squid/sources/squid/deepbook_v3.move
```diff
@@ -0,0 +1,215 @@
...
+    const SWAP_TYPE: u8 = 1;
+
+    public struct DeepbookV3SwapData has drop {
+        swap_type: u8,
+        pool_id: address,
+        has_base: bool,
+        min_output: u64,
+        base_type: String,
+        quote_type: String,
+        lot_size: u64,
```
0xaslan, line 29: This is unused, if I want to remove it, where else must I make changes?

move/squid/sources/squid/deepbook_v3.move
```diff
@@ -0,0 +1,215 @@
...
+            EWrongCoinType,
+        );
+        assert!(
+            &swap_data.quote_type == &type_name::get<Q>().into_string(),
+            EWrongCoinType,
+        );
+
+        if (swap_data.has_base) {
+            let (base_out, quote_out, _deep_required) = pool.get_quote_quantity_out(
+                self.coin_bag().get_estimate<B>(),
```
0xaslan, line 53: Should we reduce the input here by 10 bps? Since we will reduce the input amount by 10 bps later on.

move/squid/sources/squid/deepbook_v3.move
```diff
@@ -0,0 +1,215 @@
...
+            EWrongCoinType,
+        );
+
+        if (swap_data.has_base) {
+            let (_, _, deep_required) = pool.get_quote_quantity_out(
+                self.coin_bag().get_estimate<B>(),
+                clock
+            );
+            let deep_in = squid.coin_bag().get_exact_balance<DEEP>(deep_required).destroy_some().into_coin(ctx);
+            let base_in = self.coin_bag().get_balance<B>().destroy_some().into_coin(ctx);
```
0xaslan, line 107: Similarly, should we take out 10bps worth of base_in to compensate for deep_in? Would we transfer that to squid?

Foivos: That is what we discussed (Fig said maybe take a custom amount but I am now realising that anyone could craft custom transactions to use Squid's DEEP without giving them anything, so the amount held should be constant.

PR #128, opened 2024-09-06
refactor(gateway): use macro functions

npty:
# Description

This PR applies macro function to reduce duplication in `peel_*` functions in the gateway contract


move/axelar_gateway/sources/gateway.move
```diff
@@ -164,6 +168,16 @@ public fun borrow_mut(
     table::borrow_mut(&mut self.messages, command_id)
 }
 
+// -----------
+// Macros
+// -----------
```
milapsheth, line 173:
```suggestion
// ------
// Macros
// ------
```

move/axelar_gateway/sources/gateway.move
```diff
@@ -151,6 +151,10 @@ public fun setup(
     transfer::share_object(gateway);
 }
 
+// -----------
+// Index Syntax
+// -----------
+
```
milapsheth, line 157:
```suggestion
```
fine to not section it since they're already tagged by syntax(index)

move/axelar_gateway/sources/gateway.move
```diff
@@ -176,8 +190,8 @@ entry fun approve_messages(
     message_data: vector<u8>,
     proof_data: vector<u8>,
 ) {
-    let messages = peel_messages(*&message_data);
-    let proof = peel_proof(proof_data);
+    let proof = peel_data!(proof_data, |bcs| proof::peel(bcs));
```
milapsheth, line 193:
```suggestion
    let proof = peel_data!(proof_data, proof::peel);
```
does this work instead since it's already a method or does move macros require it to have the function form?

move/axelar_gateway/sources/gateway.move
```diff
@@ -164,6 +168,16 @@ public fun borrow_mut(
     table::borrow_mut(&mut self.messages, command_id)
 }
 
+// -----------
+// Macros
+// -----------
+macro fun peel_data<$T>($data: vector<u8>, $peel_fn: |&mut _| -> $T): $T {
```
milapsheth, line 174: move this to a utils package so all packages can use it?

move/axelar_gateway/sources/gateway.move
```diff
@@ -164,6 +168,16 @@ public fun borrow_mut(
...
+    let mut bcs = bcs::new($data);
+    let result = $peel_fn(&mut bcs);
```
milapsheth, line 176:
```suggestion
    let result = $T::peel(&mut bcs);
```
call peel on the type directly? If this is possible, I'd create two macro variants, one is `peel!` that always uses `T::peel` and another `peel_fn!` that uses the provided function. Might need to mention types explicitly to make `T::peel` work if it can't infer it

move/axelar_gateway/sources/gateway.move
```diff
@@ -205,8 +219,8 @@ entry fun rotate_signers(
     proof_data: vector<u8>,
     ctx: &TxContext,
 ) {
-    let weighted_signers = peel_weighted_signers(new_signers_data);
-    let proof = peel_proof(proof_data);
+    let weighted_signers = peel_data!(new_signers_data, |bcs| weighted_signers::peel(bcs)); 
+    let proof = peel_data!(proof_data, |bcs| proof::peel(bcs));
```
milapsheth, line 223:
```suggestion
    let proof: proof::Proof = peel!(proof_data);
```
would be nice if something like this can be done

move/axelar_gateway/sources/gateway.move
```diff
@@ -328,41 +342,19 @@ public fun take_approved_message(
...
-    proof
+    peel_data!(
+        message_data,
+        |bcs| {
+            let mut messages = vector::empty<Message>();
+            let mut len = bcs.peel_vec_length();
+            while (len > 0) {
+                messages.push_back(message::peel(bcs));
+                len = len - 1;
+            };
```
milapsheth, line 353:
```suggestion
            let messages = vector::tabulate(bcs.peel_vec_length(), |_| message::peel(bcs));
```
use macro funs to remove all while loops
https://blog.sui.io/move-2024-macros-beta/

move/axelar_gateway/sources/gateway.move
npty: i've tried; it doesn't work without lambda

move/axelar_gateway/sources/gateway.move
Foivos: probably not becayse $T is a type, but we need $module::peel.

npty: unfortunately, this is not possible

move/utils/sources/utils/utils.move
```diff
@@ -0,0 +1,19 @@
+/// Module: utils
+module utils::utils;
+
+use sui::bcs::{Self, BCS};
+
+/// Remaining data after BCS decoding
+const ERemainingData: u64 = 101;
+
+public macro fun peel_data<$T>(
```
milapsheth, line 9: Add section separators for consistency as in the gateway

move/utils/sources/utils/utils.move
```diff
@@ -0,0 +1,19 @@
+/// Module: utils
+module utils::utils;
+
+use sui::bcs::{Self, BCS};
+
+/// Remaining data after BCS decoding
+const ERemainingData: u64 = 101;
```
milapsheth, line 7: errors should always start from `0` within the module (i think move should show module location of the error in debugging info?)

move/utils/sources/utils/utils.move
milapsheth, line 9: Add docstring and show example usage

milapsheth, line 9:
```suggestion
public macro fun peel<$T>(
```
easier to read

move/utils/sources/utils/utils.move
```diff
@@ -0,0 +1,19 @@
...
+    $data: vector<u8>,
+    $peel_fn: |&mut BCS| -> $T,
+): $T {
+    let mut bcs = bcs::new($data);
+    let result = $peel_fn(&mut bcs);
+    assert!(bcs.into_remainder_bytes().length() == 0, error_remaining_data());
+    result
+}
+
+public fun error_remaining_data(): u64 { ERemainingData }
```
milapsheth, line 19: Why use a function for it?

mysten team recommend using this error pattern for more debugging info
```
#[error]
const ENotImplemented: vector<u8> = b"Function is not implemented";
```

Let's use that?

move/utils/sources/utils/utils.move
```diff
@@ -0,0 +1,19 @@
...
+
+public macro fun peel_data<$T>(
+    $data: vector<u8>,
+    $peel_fn: |&mut BCS| -> $T,
+): $T {
+    let mut bcs = bcs::new($data);
+    let result = $peel_fn(&mut bcs);
+    assert!(bcs.into_remainder_bytes().length() == 0, error_remaining_data());
+    result
+}
```
milapsheth, line 17: add a unit test for positive and negative case

move/axelar_gateway/sources/gateway.move
```diff
@@ -31,14 +31,16 @@ use axelar_gateway::auth::{Self, AxelarSigners, validate_proof};
...
+use axelar_gateway::weighted_signers;
 use std::ascii::String;
 use sui::address;
-use sui::bcs;
 use sui::clock::Clock;
 use sui::hash;
 use sui::table::{Self, Table};
+use utils::utils;
+#[test_only]
+use sui::bcs;
```
milapsheth, line 43: can you import this just before test only section, or does the formatter move it here?

move/axelar_gateway/sources/gateway.move
```diff
@@ -176,8 +179,8 @@ entry fun approve_messages(
     message_data: vector<u8>,
     proof_data: vector<u8>,
 ) {
-    let messages = peel_messages(*&message_data);
```
milapsheth, line 179: `*&message_data` lol

move/axelar_gateway/sources/gateway.move
```diff
@@ -186,13 +189,7 @@ entry fun approve_messages(
...
         );
 
-    let mut i = 0;
-
-    while (i < messages.length()) {
-        self.approve_message(&messages[i]);
-
-        i = i + 1;
-    };
+    messages.length().do!(|i| self.approve_message(&messages[i]));
```
milapsheth, line 192:
```suggestion
    messages.map!(|message| self.approve_message(message));
```
see other available vector macros to use based on the situation. Change `approved_message` to take in the Message itself instead of the reference

src/tx-builder.ts
```diff
@@ -271,16 +271,20 @@ export class TxBuilder {
...
+                    encoding: 'utf-8',
+                    stdio: 'pipe', // silent the output
+                }),
+            );
 
-        return { modules, dependencies, digest };
+            return { modules, dependencies, digest };
+        } catch (e) {
+            console.log(e);
+            rmTmpDir();
```
milapsheth, line 285: :+1:

test/axelar-gateway.js
```diff
@@ -81,6 +81,7 @@ describe('Axelar Gateway', () => {
             ),
         );
 
+        await publishPackage(client, deployer, 'utils');
```
milapsheth, line 84: this will change deployment scripts as well i guess?

move/axelar_gateway/sources/gateway.move
npty: this made me question a lot if there's any black magic that i don't know; I just trust the compiler 😂

move/utils/sources/utils/utils.move
npty: It will show the location of caller module instead of the location of macro function in this case. this is unlike normal functions; the macro functions will be executed the operation inline. this fact doubts me whether the error value cause any conflict with existing error code within the caller module, so i assigned the value that large enough to avoid conflict with the caller module.

move/utils/sources/utils/utils.move
npty: The macro function will be executed the operation inline, from another module in our case, that means it cannot access to the internal variable (`ERemainingData`), so i need to wrap in inside the public function.

move/utils/sources/utils/utils.move
milapsheth: hmm, i see, let's use error strings instead as recommended by the mysten team. that addresses this issue as well

src/tx-builder.ts
```diff
@@ -271,16 +271,20 @@ export class TxBuilder {
...
+            throw e;
+        }
```
milapsheth, line 287:
```suggestion
        } finally {
            rmTmpDir();
        }
```
a bit cleaner

move/axelar_gateway/sources/gateway.move
npty: In this case, i think we want to iterate over the messages vector and calls `approve_message` for each message, instead of mapping them and return a new collection of data?

move/utils/sources/utils/utils.move
milapsheth: ah, yes, macros are evaluated inline. Can you ask mysten team if this is possible somehow, or if they can add support in the future if not?

test/axelar-gateway.js
npty: We needs to add `utils` package deployment into the `deploy-contract` script. Since the `axelar_gateway` package depends on the `utils` package, we need to deploy it before the gateway contract in the test-sui workflow as well.

move/utils/sources/utils/utils.move
npty: @milapsheth As the Sui team has suggested, I've removed the error code from the `assert!` so that the error points at the macro invocation line. Additionally, to compensate the missing of error code, i've added more description about the error in the function docstring to give more details about potential error when debugging. I think this should be sufficient.

move/axelar_gateway/sources/gateway.move
milapsheth: Looks like they have `foreach` macro, use that instead

move/utils/sources/utils/utils.move
```diff
@@ -0,0 +1,52 @@
...
+
+use sui::bcs::{Self, BCS};
+
+// -----
+// Macros
+// -----
+
+/**
+* Peel data from a BCS encoded vector
+* @dev This macro is used to peel data from a BCS encoded vector
```
milapsheth, line 12: Don't think natspec format should be used. Use the same as other move code / Rust
https://github.com/MystenLabs/sui/blob/db5ba7aa9564af91de0aa9ad58ac0f2c9f86eac2/crates/sui-framework/packages/sui-framework/sources/kiosk/kiosk_extension.move#L62

move/utils/sources/utils/utils.move
```diff
@@ -0,0 +1,52 @@
...
+public macro fun peel<$T>($data: vector<u8>, $peel_fn: |&mut BCS| -> $T): $T {
+    let mut bcs = bcs::new($data);
+    let result = $peel_fn(&mut bcs);
+    assert!(bcs.into_remainder_bytes().length() == 0);
+    result
+}
+
+
+#[test]
+fun test_peel_success() {
```
milapsheth, line 34: let's avoid prefixing with `test_*` similar to rust, since `#[test]` already implies it. Instead we can make it descriptive in terms of what it does. See our cosmwasm examples [here](https://github.com/axelarnetwork/axelar-amplifier/blob/main/contracts/axelarnet-gateway/tests/execute.rs#L19)
```suggestion
fun peel_bcs_data_succeeds() {
```

Update other test names as well as you refactor

move/utils/sources/utils/utils.move
```diff
@@ -0,0 +1,52 @@
...
+    let peeled_data: vector<u8> = peel!(
+        data,
+        |bcs| bcs::peel_vec_u8(bcs),
+    );
+    assert!(peeled_data == test_bytes, 0);
+}
+
+#[test]
+#[expected_failure]
+fun test_peel_error() {
```
milapsheth, line 46:
```suggestion
fun peel_invalid_bcs_fails() {
```

move/utils/sources/utils/utils.move
```diff
@@ -0,0 +1,52 @@
...
+
+#[test]
+fun test_peel_success() {
+    let test_bytes = b"test";
+    let data = bcs::to_bytes(&test_bytes);
+    let peeled_data: vector<u8> = peel!(
+        data,
+        |bcs| bcs::peel_vec_u8(bcs),
+    );
+    assert!(peeled_data == test_bytes, 0);
```
milapsheth, line 41:
```suggestion
    assert!(peeled_data == test_bytes);
```
with the current sui version does the failing print more info, like the line number? If yes, then we can avoid the abort codes

src/tx-builder.ts
```diff
@@ -271,16 +271,21 @@ export class TxBuilder {
...
+                execSync(`sui move build --dump-bytecode-as-base64 --path ${path.join(moveDir, packageName)} --install-dir ${tmpdir}`, {
+                    encoding: 'utf-8',
+                    stdio: 'pipe', // silent the output
+                }),
+            );
 
-        return { modules, dependencies, digest };
+            return { modules, dependencies, digest };
+        } catch (e) {
+            console.log(e);
```
milapsheth, line 284: should be the caller's choice whether to log it or not (if the error isn't handled then it'll be shown to the user anyways

milapsheth (approved):

move/axelar_gateway/sources/gateway.move
npty: i don't see `foreach` macro in [their doc](https://blog.sui.io/move-2024-macros-beta/). have they mentioned that anywhere?

PR #135, opened 2024-09-11
feat: version control

Foivos:


move/version_control/sources/version_control.move
```diff
@@ -0,0 +1,136 @@
...
+// -----
+public struct VersionControl has store, copy, drop {
+    allowed_functions: vector<VecSet<vector<u8>>>,
+}
+
+// ----------------
+// Public Functions
+// ----------------
+
+/// Create a new Version Controll object by passing raw allowed_functions data.
```
milapsheth, line 23:
```suggestion
/// Create a new Version Control object by passing in the allowed functions per version.
```

move/version_control/sources/version_control.move
```diff
@@ -0,0 +1,136 @@
...
+public fun push_back(self: &mut VersionControl, function_names: vector<vector<u8>>) {
+    self.allowed_functions.push_back(
+        vec_set::from_keys(
+            function_names
+        )
+    );
+}
+
+/// Call this at the begining of each version controlled function. For example
+/// public fun do_something(data: &mut DataType) {
```
milapsheth, line 57: wrap inside a code block

move/version_control/sources/version_control.move
```diff
@@ -0,0 +1,136 @@
...
+
+/// Create a new Version Controll object by passing raw allowed_functions data.
+/// You are supposed to pass a vector of the bytes of the functions that are allowed per version. For example:
+/// vector [
+///     vector [ b"v0_function" ],
+///     vector [ b"v0_function", b"v1_function"],
+/// ]
+/// Would allow only `v0_function` to be called on version == 0, and both `v0_function` and `v1_function` to be called on version == 1.
+/// This is done to simplify the instantiation syntax of VersionControl.
+public fun new(allowed_functions: vector<vector<vector<u8>>>): VersionControl {
```
milapsheth, line 31:
```suggestion
public fun new(allowed_functions: vector<vector<String>>): VersionControl {
```
I forget if there was an issue with using strings instead. If not, let's use ascii string

move/version_control/sources/version_control.move
```diff
@@ -0,0 +1,136 @@
+/// This module implements a custom version control scheme to maximize versioning customizability.
+module version_control::version_control;
+
+use sui::vec_set::{Self, VecSet};
+
+// ------
+// Errors
+// ------
+/// This is thrown when the function is not supported in the version passed on `check`
+const EFunctionNotSupported: u64 = 0;
```
milapsheth, line 10:
```suggestion
#[error]
const EFunctionNotSupported: vector<u8> = b"function is not supported in this version";
```
use the new error syntax

milapsheth (approved):

move/version_control/sources/version_control.move
Foivos: There is no clean way to input strings, so the instantiations looks messier with strings, and they will be hardcoded, not input, so there is no benefit in storing as String

milapsheth: the input can be bytes, but it can be converted into string before storing in the object for better display in explorers.

move/version_control/sources/version_control.move
```diff
@@ -0,0 +1,140 @@
...
+///     vector [ b"v0_function", b"v1_function"],
+/// ]
+/// ```
+/// Would allow only `v0_function` to be called on version == 0, and both `v0_function` and `v1_function` to be called on version == 1.
+/// This is done to simplify the instantiation syntax of VersionControl.
+public fun new(allowed_functions: vector<vector<vector<u8>>>): VersionControl {
+    VersionControl {
+        allowed_functions: allowed_functions.map!(
+            |function_names| vec_set::from_keys(
+                function_names
```
milapsheth, line 37:
```suggestion
                function_names.map!(|function_name| ascii::String(function_name))
```
improve display for the object

PR #140, opened 2024-09-12
refactor(auth): extract validation to weight signers and proof

npty:
# Description

https://axelarnetwork.atlassian.net/browse/AXE-4845


move/axelar_gateway/sources/types/weighted_signers.move
```diff
@@ -16,21 +16,63 @@ public struct WeightedSigners has copy, drop, store {
...
+/// 1. The signers are in ascending order by their public key.
+/// 2. The threshold is greater than zero.
+/// 3. The threshold is less than or equal to the total weight of the signers.
+public(package) fun validate(self: &WeightedSigners) {
+    let signers = self.signers();
+    assert!(!vector::is_empty(&signers), EInvalidOperators);
+
+    validate_signers_order(signers);
+    let total_weight = calculate_total_weight(signers);
+    validate_threshold(self.threshold(), total_weight);
```
npty, line 63: This section trades off some performance. Previously, we verified everything in a single loop.

move/axelar_gateway/sources/types/proof.move
```diff
@@ -60,6 +61,33 @@ public(package) fun recover_pub_key(
...
+    message: vector<u8>
+) {
+    let signers = self.signers();
+    let signatures = self.signatures();
+    assert!(signatures.length() != 0, ELowSignaturesWeight);
+
+    let threshold = signers.threshold();
+    let mut total_weight: u128 = 0;
+
+    signatures.do_ref!<Signature>(|signature| {
```
milapsheth, line 78:
```suggestion
    signatures.do_ref!(|signature| {
```
not needed since it should be inferred

move/axelar_gateway/sources/types/proof.move
```diff
@@ -60,6 +61,33 @@ public(package) fun recover_pub_key(
...
+        let pub_key = signature.recover_pub_key(&message);
+
+        let weight = signers.find_signer_weight(&pub_key);
```
milapsheth, line 81: This makes it asymptotically worse. Find would need to start from the next index but Move might not allow doing that cleanly.

move/axelar_gateway/sources/types/weighted_signers.move
```diff
@@ -16,21 +16,63 @@ public struct WeightedSigners has copy, drop, store {
...
     }
 }
 
+/// Validates the weighted signers. The following must be true:
+/// 1. The signers are in ascending order by their public key.
+/// 2. The threshold is greater than zero.
+/// 3. The threshold is less than or equal to the total weight of the signers.
+public(package) fun validate(self: &WeightedSigners) {
+    let signers = self.signers();
+    assert!(!vector::is_empty(&signers), EInvalidOperators);
```
milapsheth, line 59:
```suggestion
    assert!(!self.signers.is_empty(), EInvalidOperators);
```
struct fields can be accessed directly within the same module

milapsheth: change the error name to `EInvalidSigners`

move/axelar_gateway/sources/types/weighted_signers.move
```diff
@@ -16,21 +16,63 @@ public struct WeightedSigners has copy, drop, store {
...
+
+    validate_signers_order(signers);
```
milapsheth, line 61:
```suggestion
    self.validate_signers();
```
move the length check and order check under this method, instead of copying the field data

move/axelar_gateway/sources/types/weighted_signers.move
milapsheth: Can you compare gas usage before and after this change on their testnet?

move/axelar_gateway/sources/types/weighted_signers.move
```diff
@@ -16,21 +16,63 @@ public struct WeightedSigners has copy, drop, store {
...
+    let total_weight = calculate_total_weight(signers);
```
milapsheth, line 62: change this to `self.validate_threshold()`. it can internally sum and check.

move/axelar_gateway/sources/types/weighted_signer.move
```diff
@@ -31,6 +31,9 @@ public fun weight(self: &WeightedSigner): u128 {
 // ------
 
 const EInvalidPubKeyLength: u64 = 0;
+const EInvalidOperators: u64 = 1;
```
milapsheth, line 34:
```suggestion
const EInvalidSigner: u64 = 1;
```

move/axelar_gateway/sources/types/weighted_signers.move
```diff
@@ -47,6 +89,38 @@ public(package) fun nonce(self: &WeightedSigners): Bytes32 {
...
+/// -----
+
+/// Validates the order of the signers.
+/// The signers must be in ascending order by their public key.
+/// Otherwise, the error `EInvalidOperators` is raised.
+fun validate_signers_order(signers: vector<WeightedSigner>) {
+    let mut previous = weighted_signer::default();
+    signers.do!<WeightedSigner>(
+        |signer| {
+            signer.validate(&previous);
```
milapsheth, line 103: `validate` method should only take `self` instead of another signer. Use `lt` comparison separately to check. I'm also fine with just checking the weight here explicitly.

move/axelar_gateway/sources/types/weighted_signers.move
```diff
@@ -47,6 +89,38 @@ public(package) fun nonce(self: &WeightedSigners): Bytes32 {
...
+/// -----
+/// Internal Functions
+/// -----
+
+/// Validates the order of the signers.
+/// The signers must be in ascending order by their public key.
+/// Otherwise, the error `EInvalidOperators` is raised.
+fun validate_signers_order(signers: vector<WeightedSigner>) {
+    let mut previous = weighted_signer::default();
+    signers.do!<WeightedSigner>(
```
milapsheth, line 101:
```suggestion
    signers.do!(
```

move/axelar_gateway/sources/types/proof.move
milapsheth: `fold` could be used but I don't see it being readable. I prefer sticking with the while loop

move/axelar_gateway/sources/auth.move
```diff
@@ -102,13 +95,13 @@ public(package) fun validate_proof(
...
+    let message = proof::new_message_to_sign(
+        self.domain_separator,
         signers_hash,
-        data_hash,
-    };
+        data_hash
+    );
 
-    proof.validate(bcs::to_bytes(&message));
+    proof.validate(message);
```
milapsheth, line 104: this is a bit weird. MessageToSign is defined in proof, and then returned here but then immediately passed back to the proof module and then encoded. I'd preserve the MessageToSign definition in this module and pass in the bcs encoded message `proof` module shouldn't care about the `MessageToSign` definition. It should be able to validate a proof over an arbitrary message. It's the `auth` module that encodes the message in a certain way

move/axelar_gateway/sources/types/weighted_signers.move
```diff
@@ -55,21 +72,21 @@ public(package) fun peel(bcs: &mut BCS): WeightedSigners {
...
 public(package) fun find_signer_weight(
     self: &WeightedSigners,
+    signer_index: u64,
     pub_key: &vector<u8>,
-): u128 {
-    let mut signer = self.find!(|signer| signer.pub_key() == pub_key);
+): (u128, u64) {
+    // Find the signer by its public key
+    // The index of the signer is returned as well for reuse in the macro call
+    let (signer, index) = self.find!(signer_index, |signer| signer.pub_key() == pub_key);
```
milapsheth, line 87: Don't think we need a macro here. It's not being reused anywhere else so it's just adding more complexity

move/axelar_gateway/sources/types/proof.move
```diff
@@ -66,19 +84,23 @@ public(package) fun recover_pub_key(
...
     let signers = self.signers();
     let signatures = self.signatures();
     assert!(signatures.length() != 0, ELowSignaturesWeight);
 
     let threshold = signers.threshold();
     let mut total_weight: u128 = 0;
+    let mut signer_index = 0;
+    let message_bytes = bcs::to_bytes(&message);
 
     signatures.do_ref!(|signature| {
```
milapsheth, line 98: i'd convert this back to a while loop. it's not very functional due to the mutation and the return in the middle, so i find it harder to read

move/axelar_gateway/sources/types/proof.move
```diff
@@ -66,19 +84,23 @@ public(package) fun recover_pub_key(
...
-        let pub_key = signature.recover_pub_key(&message);
+        let pub_key = signature.recover_pub_key(&message_bytes);
+
+        let (weight, index) = signers.find_signer_weight(signer_index, &pub_key);
```
milapsheth, line 101: `find_signer_weight` -> `find_weight_by_pubkey_from` maybe. I'd also define it in this module. It's too specific to the logic here to exist in the other module

move/axelar_gateway/sources/types/weighted_signer.move
```diff
@@ -31,9 +31,7 @@ public fun weight(self: &WeightedSigner): u128 {
 // ------
 
 const EInvalidPubKeyLength: u64 = 0;
-const EInvalidSigners: u64 = 1;
-const EInvalidWeights: u64 = 2;
-const EMalformedSigners: u64 = 3;
+const EInvalidWeights: u64 = 1;
```
milapsheth, line 34:
```suggestion
const EInvalidWeight: u64 = 1;
```

move/axelar_gateway/sources/types/weighted_signers.move
```diff
@@ -92,22 +109,25 @@ public(package) fun nonce(self: &WeightedSigners): Bytes32 {
 /// Internal Functions
 /// -----
 
-/// Validates the order of the signers.
+/// Validates the order of the signers and the length of the signers.
 /// The signers must be in ascending order by their public key.
 /// Otherwise, the error `EInvalidSigners` is raised.
-fun validate_signers_order(signers: vector<WeightedSigner>) {
+fun validate_signers(self: &WeightedSigners) {
+    let signers = self.signers;
```
milapsheth, line 116: this is creating a copy instead of a reference

milapsheth, line 116:
```suggestion
    let signers = &self.signers;
```
or use self.signers directly, probably cleaner

move/axelar_gateway/sources/types/weighted_signers.move
```diff
@@ -116,7 +136,9 @@ fun calculate_total_weight(signers: vector<WeightedSigner>): u128 {
 /// Validates the threshold.
 /// The threshold must be greater than zero and less than or equal to the total weight of the signers.
 /// Otherwise, the error `EInvalidThreshold` is raised.
-fun validate_threshold(threshold: u128, total_weight: u128) {
+fun validate_threshold(self: &WeightedSigners) {
+    let threshold = self.threshold();
+    let total_weight = self.total_weight();
     assert!(threshold != 0 && total_weight >= threshold, EInvalidThreshold);
```
milapsheth, line 142:
```suggestion
    assert!(self.threshold != 0 && self.total_weight >= self.threshold, EInvalidThreshold);
```
this was creating copies of the fields unnecessarily

move/axelar_gateway/sources/types/weighted_signers.move
```diff
@@ -92,22 +109,25 @@ public(package) fun nonce(self: &WeightedSigners): Bytes32 {
...
 /// -----
 
-/// Validates the order of the signers.
+/// Validates the order of the signers and the length of the signers.
 /// The signers must be in ascending order by their public key.
 /// Otherwise, the error `EInvalidSigners` is raised.
-fun validate_signers_order(signers: vector<WeightedSigner>) {
+fun validate_signers(self: &WeightedSigners) {
+    let signers = self.signers;
+    assert!(!signers.is_empty(), EInvalidSigners);
```
milapsheth, line 117:
```suggestion
    assert!(!self.signers.is_empty(), EInvalidSigners);
```

move/axelar_gateway/sources/types/weighted_signers.move
```diff
@@ -92,22 +109,25 @@ public(package) fun nonce(self: &WeightedSigners): Bytes32 {
...
     let mut previous = weighted_signer::default();
```
milapsheth, line 118:
```suggestion
    let mut previous = &weighted_signer::default();
```

move/axelar_gateway/sources/types/weighted_signers.move
```diff
@@ -92,22 +109,25 @@ public(package) fun nonce(self: &WeightedSigners): Bytes32 {
...
     signers.do!(
         |signer| {
-            signer.validate(&previous);
+            signer.validate();
+            assert!(previous.lt(&signer), EInvalidSigners);
```
milapsheth, line 122:
```suggestion
            signer.validate();
            assert!(previous.lt(signer), EInvalidSigners);
```

move/axelar_gateway/sources/types/weighted_signers.move
```diff
@@ -16,21 +16,79 @@ public struct WeightedSigners has copy, drop, store {
...
+public(package) fun find_signer_weight(
+    self: &WeightedSigners,
+    signer_index: u64,
+    pub_key: &vector<u8>,
+): (u128, u64) {
+    // Find the signer by its public key
+    // The index of the signer is returned as well for reuse in the macro call
+    let (signer, index) = self.find!(signer_index, |signer| signer.pub_key() == pub_key);
+
+    (signer.parse_weight(), index)
```
milapsheth, line 89: weight doesn't need to be validated again since it was already checked for being non-zero during signer rotation

move/axelar_gateway/sources/types/weighted_signers.move
```diff
@@ -47,6 +105,43 @@ public(package) fun nonce(self: &WeightedSigners): Bytes32 {
...
+/// -----
+
+/// Validates the order of the signers and the length of the signers.
+/// The signers must be in ascending order by their public key.
+/// Otherwise, the error `EInvalidSigners` is raised.
+fun validate_signers(self: &WeightedSigners) {
+    let signers = self.signers;
+    assert!(!signers.is_empty(), EInvalidSigners);
+    let mut previous = weighted_signer::default();
+    signers.do!(
```
milapsheth, line 119:
```suggestion
    signers.do_ref!(
```

move/axelar_gateway/sources/types/proof.move
```diff
@@ -60,6 +61,62 @@ public(package) fun recover_pub_key(
...
+        signer_index = index;
+
+        total_weight = total_weight + weight;
+
+        i = i + 1;
+
+        if (total_weight >= threshold) return
+    };
+
+    assert!(total_weight >= threshold, ELowSignaturesWeight);
```
milapsheth, line 96:
```suggestion
    abort ELowSignaturesWeight;
```
if this is reached, it's always insufficient. Double check the implementation matches the EVM gateway for sanity
https://github.com/axelarnetwork/axelar-gmp-sdk-solidity/blob/main/contracts/governance/BaseWeightedMultisig.sol

move/axelar_gateway/sources/types/proof.move
```diff
@@ -60,6 +61,62 @@ public(package) fun recover_pub_key(
...
+}
+
+/// Finds the weight of a signer in the weighted signers by its public key.
+public(package) fun find_weight_by_pub_key_from(
```
milapsheth, line 100:
```suggestion
fun find_weight_by_pub_key_from(
```
should be internal

move/axelar_gateway/sources/types/proof.move
```diff
@@ -60,6 +61,62 @@ public(package) fun recover_pub_key(
...
+    weight_signers: &WeightedSigners,
+    signer_index: u64,
```
milapsheth, line 102:
```suggestion
    signers: &WeightedSigners,
    signer_index: u64,
```
or `weighted_signers`

move/axelar_gateway/sources/types/proof.move
```diff
@@ -60,6 +61,62 @@ public(package) fun recover_pub_key(
...
+    pub_key: &vector<u8>,
+): (u128, u64) {
+    let signers = weight_signers.signers();
```
milapsheth, line 105: Make this return a reference perhaps so it doesn't copy the whole vector

move/axelar_gateway/sources/types/proof.move
npty: i've addressed this implementation to match with the EVM gateway, but i'm not sure about [this check](https://github.com/axelarnetwork/axelar-gmp-sdk-solidity/blob/febcd9fb2b4c0a6db23f5265f3ae9e115a3f2fc6/contracts/governance/BaseWeightedMultisig.sol#L219-L224) because it doesn't include in our [original implementation](https://github.com/axelarnetwork/axelar-cgp-sui/blob/479ddb615520f42eb25f7447522d1d1baa3cee94/move/axelar_gateway/sources/auth.move#L155-L192) as well.

milapsheth: Yeah, that was an optional check added from a recent audit. Might consider removing it since it's overly restrictive, you can ignore it.

move/axelar_gateway/sources/types/proof.move
```diff
@@ -64,45 +65,46 @@ public(package) fun recover_pub_key(
...
 
-        let (weight, index) = find_weight_by_pub_key_from(signers, signer_index, &pub_key);
+        let (weight, index) = find_weight_by_pub_key_from(
+            signers,
+            signer_index,
+            &pub_key,
+        );
 
-        signer_index = index;
+        assert!(signer_index < signatures_length, EMalformedSigners);
```
milapsheth, line 88: this is being done in the [find](https://github.com/axelarnetwork/axelar-cgp-sui/pull/140/files/e406dfb6fa0b8faa17d32d99cd45bc799ee88c49..37861f3761282d252ac73e3aecc7b7c6a498d2df#diff-92469ab1c1944669420bbcbf13715f8cbc924ce9ff1a0b2de4da86906a67a1e6R117) helper method

milapsheth (approved):

npty:
I've tried comparing gas usage before and after this changes by rotating the signers with the scripts in `axelar-contract-deployments`. 

**My Setup**
- Generate 70 public keys with [this script](https://gist.github.com/npty/35ec1600440336d9ef8b7e6d3483414d)
- Override the signers with hardcoded generated public keys above [here](https://github.com/axelarnetwork/axelar-contract-deployments/blob/bdcbe03fa9aaaabadd63d5bbfae30c7f1196041c/sui/gateway.js#L252) 
- Set the threshold to 35
- Deploy and run rotate signers script as usual 

I've tested 2 times, each time I deployed the gateway contract separately, just to make sure everything is in clean state.

# Before

- https://suiscan.xyz/testnet/tx/Aw6Q8z6Suis61aaJY1Euj34YgA1G29s9Wz5kAE4HCVhN 
- https://suiscan.xyz/testnet/tx/HHBKA5kC8zq8oKXTeQiCo4JpDjTg3CHmbEZ8H5CAmEwf 

Both Gas Usage: 0.62 SUI

# After

- https://suiscan.xyz/testnet/tx/Afx5c4rwchveTcyR1ifXy9dC75mUNTGSkumC2qCQUQDL 
- https://suiscan.xyz/testnet/tx/M1d6qLkFTJcVxZ3mcWVc4ws2hKhf8oNh5Ypys2CEbyH   

Both Gas Usage: 0.38 SUI

After the refactoring, the gas usage seems to be reduced.

PR #147, opened 2024-09-19
refactor: clean up updateMoveToml function

npty:
# Description

This PR simplified complex implementation due to toml file parsing with https://github.com/iarna/iarna-toml

https://axelarnetwork.atlassian.net/browse/AXE-4976


package.json
```diff
@@ -40,6 +40,7 @@
   },
   "dependencies": {
     "@cosmjs/cosmwasm-stargate": "^0.32.2",
+    "@iarna/toml": "^2.2.5",
```
milapsheth, line 43: this doesn't seem to be maintained anymore, how about this [package](https://www.npmjs.com/package/smol-toml) instead? not a huge issue since it's just toml parsing

src/utils.ts
```diff
@@ -1,5 +1,6 @@
 import { execSync } from 'child_process';
 import fs from 'fs';
+import toml from '@iarna/toml';
 import { getFullnodeUrl } from '@mysten/sui/client';
 
 export function getModuleNameFromSymbol(symbol: string) {
```
milapsheth, line 6: this method seems unnecessarily complicated as well, can you take a look? where is this used?

test/utils.test.js
milapsheth, line 1: `.test` in the file name is unnecessary since it's inside the `test` dir

.gitignore
```diff
@@ -75,3 +75,6 @@ move_compile/
 
 # binaries
 sui-binaries/
+
+# move test folder
+move-test/
```
milapsheth, line 80:
```suggestion
test/move
```
keep it inside the test folder for clarity?

test/utils.test.js
```diff
@@ -0,0 +1,24 @@
...
+        copyMovePackage(testPackageName, undefined, moveTestDir);
+
+        // Update the Move.toml file for the test package
+        updateMoveToml(testPackageName, testPackageId, moveTestDir);
+
+        const moveToml = toml.parse(fs.readFileSync(`${moveTestDir}/${testPackageName}/Move.toml`, 'utf8'));
+
+        expect(moveToml.package.published_at).to.equal(testPackageId);
+        expect(moveToml.addresses[testPackageName]).to.equal(testPackageId);
+    });
```
milapsheth, line 23: add an `after` block to delete the move test dir after the test runs

test/utils.test.js
```diff
@@ -0,0 +1,24 @@
+const { expect } = require('chai');
+const toml = require('@iarna/toml');
+const fs = require('fs');
+const { updateMoveToml, copyMovePackage } = require('../dist/utils');
+
+describe('UpdateMoveToml', () => {
```
milapsheth, line 6:
```suggestion
describe('Utils', () => {
```

test/utils.test.js
```diff
@@ -0,0 +1,24 @@
...
+    const moveTestDir = `${__dirname}/../move-test`;
+
+    it('should insert published-at and address fields correctly', () => {
```
milapsheth, line 9: wrap this in another describe block called `updateMoveToml` for better nesting

milapsheth (approved):

test/utils.test.js
npty: there's a `test/utils.js` which doesn't seem like a test file, so i add `.test` to other test files to separate them out.

package.json
npty: will try

test/utils.test.js
npty: updated in another pr

test/utils.test.js
npty: same to above

src/utils.ts
npty: seems like it doesn't use anywhere. do you really need it?

milapsheth: don't see it used in deployments repo as well, let's remove it

test/utils.test.js
milapsheth: hmm, let's rename that instead, maybe `common.js`, or `testutils.js`?

milapsheth (approved):

PR #154, opened 2024-09-30
test: check exported structs and public functions

milapsheth:
[AXE-5293]

This integration test checks for any changes to the public interface to any of the Move packages to detect breaking behaviour. After a production release of the contract, public interfaces aren't allowed to be changed by Move as required by package upgrades. So, any changes to testdata should be reviewed carefully. Furthermore, this test also serves as a quick way to check what methods are being exported by each module, and in case visibility should be restricted.

Testdata can be updated by running `GOLDEN_TESTS=true npm run test-js`.

[AXE-5293]: https://axelarnetwork.atlassian.net/browse/AXE-5293?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


test/testdata/interface_axelar_gateway_gateway.json
```diff
@@ -0,0 +1,82 @@
...
+  },
+  "publicFunctions": {
+    "prepare_message": {
+      "name": "prepare_message",
+      "visibility": "public",
+      "params": {
+        "channel#0#0": "&Channel",
+        "destination_chain#0#0": "String",
+        "destination_address#0#0": "String",
+        "payload#0#0": "vector<u8>"
```
npty, line 41: just wondering: why do params have a `#0#0` postfix here? could we remove it?

milapsheth: no idea, i'd like to know in case it changes for some reason so i don't want to remove it.

npty (approved):

PR #163, opened 2024-10-04
feat(its)!: use versioned for its and some renaming

Foivos:


move/its/sources/its.move
```diff
@@ -1,287 +1,558 @@
...
+        balance,
+        source_address,
+        destination_chain,
+        destination_address,
+        metadata,
+        version,
+    ) = ticket.destroy();
+    assert!(version <= VERSION, ENewerTicket);
+
+    let value = self.value_mut!(b"send_interchain_transfer");
```
milapsheth, line 209: same

move/its/sources/its.move
```diff
@@ -1,287 +1,558 @@
...
+    self: &mut ITS,
+    coin_info: CoinInfo<T>,
+    coin_management: CoinManagement<T>,
+): TokenId {
+    let token_id = token_id::from_coin_data(&coin_info, &coin_management);
 
-public fun get_coin_info<T>(self: &ITS, token_id: TokenId): &CoinInfo<T> {
-    &get_coin_data<T>(self, token_id).coin_info
-}
+    self.value_mut!(b"register_coin").add_registered_coin(token_id, coin_management, coin_info);
```
milapsheth, line 140: always check this first in the method for consistency. Makes it easier to audit

move/its/sources/its.move
```diff
@@ -1,287 +1,558 @@
...
+    let remote_decimals = (reader.read_u256() as u8);
+    let distributor_bytes = reader.read_bytes();
+    let decimals = if (remote_decimals > DECIMALS_CAP) DECIMALS_CAP
+    else remote_decimals;
+    let (treasury_cap, mut coin_metadata) = value.remove_unregistered_coin<T>(
+        token_id::unregistered_token_id(&symbol, decimals),
+    );
+
+    treasury_cap.update_name(&mut coin_metadata, name);
+    //coin::update_symbol(&treasury_cap, &mut coin_metadata, symbol);
```
milapsheth, line 320: commented code?

move/its/sources/its.move
```diff
@@ -1,287 +1,558 @@
...
-    while (i < length) {
-        self.set_trusted_address(
-            ascii::string(chain_names.pop_back()),
-            ascii::string(trusted_addresses.pop_back()),
-        );
-        i = i + 1;
-    }
+    let value = self.value_mut!(b"receive_deploy_interchain_token");
+
+    let (_, payload) = decode_approved_message(value, approved_message);
```
milapsheth, line 301: we're differentiating from the gateway here by keeping most of the logic in this contract vs the versioned module. Shall we move it there for consistency? While the coupling is relatively lower, it still feels a bit awkward to keep them separate?

move/its/sources/its.move
```diff
@@ -1,287 +1,558 @@
...
-    self.unregistered_coin_types.add(token_id, type_name);
+    let value = self.value_mut!(b"set_trusted_addresses");
+    
+    let (
+        source_chain,
+        _,
+        source_address,
+        payload,
+    ) = channel::consume_approved_message(
+        value.channel(),
```
npty, line 430: should be `value.channel_mut()` as previously defined in `service.move`?

move/its/sources/discovery.move
```diff
@@ -79,31 +78,32 @@ public fun get_call_info(self: &ITS, mut payload: vector<u8>): Transaction {
...
-    self: &ITS,
+    its: &ITS,
     reader: &mut AbiReader,
 ): Transaction {
     let token_id = token_id::from_u256(reader.read_u256());
     reader.skip_slot(); // skip source_address
     let destination_address = address::from_bytes(reader.read_bytes());
     reader.skip_slot(); // skip amount
     let data = reader.read_bytes();
+    let value = its.package_value();
```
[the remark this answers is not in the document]
Foivos: I like keeping the logic in this file since it is telling relayer which public functions to call, and it has nothing to do with actual working logic.

move/its/sources/its.move
Foivos: removed

move/its/sources/its.move
Foivos: I like it better this way because the storage logic lives in `its_v0` but the public facing logic live in `its`. I can change it if you think it's too awkward.

PR #165, opened 2024-10-08
fix: its example for versioned ITS

npty:
# Description

This PR fixes test for the new versioned ITS. 

## Changes
- Most changes are due to `service.move` was now merged to `its.move`, so any move calls should be targetted on `its`, instead of `service`. 
- Another change is to update a way to get channelId from the `ITS_V0` object instead of `ITS` object

### DIFF
--- a/test/its.js
+++ b/test/its.js
@@ -28,7 +28,7 @@ const { ITSMessageType } = require('../dist/types');
 const { TxBuilder } = require('../dist/tx-builder');
 const { keccak256, defaultAbiCoder, toUtf8Bytes, hexlify, randomBytes } = require('ethers/lib/utils');
 
-describe.only('ITS', () => {
+describe('ITS', () => {
     // Sui Client
     let client;
     const network = process.env.NETWORK || 'localnet';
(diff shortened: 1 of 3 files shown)


milapsheth (approved):

test/its.js
```diff
@@ -116,7 +117,8 @@ describe('ITS', () => {
             singleton: findObjectId(deployments.example.publishTxn, 'its::Singleton'),
             tokenTreasuryCap: findObjectId(deployments.example.publishTxn, 'TreasuryCap'),
             tokenCoinMetadata: findObjectId(deployments.example.publishTxn, 'CoinMetadata'),
-            its: findObjectId(deployments.its.publishTxn, 'ITS'),
+            its: findObjectId(deployments.its.publishTxn, 'its::ITS'),
+            itsv0: findObjectId(deployments.its.publishTxn, 'its_v0::ITS_V0'),
```
milapsheth, line 121:
```suggestion
            its_v0: findObjectId(deployments.its.publishTxn, 'its_v0::ITS_V0'),
```

milapsheth (approved):

test/its.js
npty: perhaps camel case like other objectIds `itsV0`?

milapsheth: yeah, `itsV0` sounds good

milapsheth (approved):

PR #189, opened 2024-10-21
chore: split builds for commonjs, esm, and web

npty:
# Description

https://axelarnetwork.atlassian.net/browse/AXE-6088

This PR splits build for `esm`, `commonjs`, and `web` separately. the functions that requires node.js environment are completely removed from the web package. the node.js environment packages are splitted into two packages: `esm`, and `cjs`. the `tx-builder` class is now splitted into two versions: node.js version and web version. the web version removes some functionalities such as `publishInterchainToken` and `publishPackage`.

here's the structure of the `src` folder:

```
src
├── common
│   ├── bcs.ts
│   ├── execute.ts
│   ├── tx-builder-base.ts
│   ├── types.ts
│   └── utils.ts
├── node
│   ├── index.ts
│   ├── node-utils.ts
│   └── tx-builder.ts
└── web
   └── index.ts
```

the tsconfig files are defined separately for each build which are used in the build scripts inside `package.json`. The build scripts also have a post-processing script which flattens the file structure and fixes the import paths.


package.json
```diff
@@ -12,10 +12,35 @@
...
+      "types": "./dist/cjs/index.d.ts"
+    }
+  },
   "scripts": {
     "build-move": "./scripts/run.sh build",
-    "build-ts": "tsc",
+    "build-ts": "npm run build-ts:cjs && npm run build-ts:esm && npm run build-ts:web",
+    "build-ts:cjs": "tsc -p tsconfig.cjs.json && node scripts/flatten.js dist/cjs",
+    "build-ts:esm": "tsc -p tsconfig.esm.json && node scripts/flatten.js dist/esm",
+    "build-ts:web": "tsc -p tsconfig.web.json && node scripts/flatten.js dist/web",
```
canhtrinh, line 43: consider also invoking here the `verify-web-build` script as a post-build step

.github/workflows/verify-build.yaml
```diff
@@ -0,0 +1,26 @@
+name: Verify Build
```
canhtrinh, line 1: not opposed to it, but is this action necessary? if you validate on build, you won't need to invoke this on the fly with PRs.

npty: yes, we can verify it after web build script as well. already moved there.

canhtrinh (approved):

Foivos (approved):

PR #202, opened 2024-11-15
feat(its)!: make event for interchain transfer consistent with EVM

Foivos:
[AXE-6579]

[AXE-6579]: https://axelarnetwork.atlassian.net/browse/AXE-6579?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/move/its/sources/events.move
+++ b/move/its/sources/events.move
@@ -136,6 +136,8 @@ public(package) fun unregistered_coin_received<T>(
 use its::coin::COIN;
 #[test_only]
 use its::token_id;
+#[test_only]
+use utils::utils;
 
 // -----
 // Tests
@@ -168,9 +170,7 @@ fun test_interchain_transfer() {
         amount,
         &data2,
     );
-    let events = event::events_by_type<InterchainTransfer<COIN>>();
-
-    assert!(events.length() == 2);
+    let events = utils::multiple_events<InterchainTransfer<COIN>>(2);
 
     assert!(events[0].data_hash == data1_hash);
     assert!(events[1].data_hash == data2_hash);
--- a/move/its/sources/events.move
+++ b/move/its/sources/events.move
@@ -202,7 +202,7 @@ fun test_interchain_transfer_empty_data() {
         amount,
         &data,
     );
-    let event = utils::single_event<InterchainTransfer<COIN>>();
+    let event = utils::assert_single_event<InterchainTransfer<COIN>>();
 
     assert!(event.data_hash == data_hash);
     assert!(event.source_address == source_address);
@@ -230,7 +230,7 @@ fun test_interchain_transfer_nonempty_data() {
         amount,
         &data,
     );
-    let event = utils::single_event<InterchainTransfer<COIN>>();
+    let event = utils::assert_single_event<InterchainTransfer<COIN>>();
 
     assert!(event.data_hash == data_hash);
     assert!(event.source_address == source_address);
--- a/move/its/sources/events.move
+++ b/move/its/sources/events.move
@@ -48,6 +48,11 @@ public struct UnregisteredCoinReceived<phantom T> has copy, drop {
     decimals: u8,
 }
 
+public struct FlowLimitSet<phantom T> has copy, drop {
+    token_id: TokenId,
+    flow_limit: u64,
+}
+
 // -----------------
 // Package Functions
 // -----------------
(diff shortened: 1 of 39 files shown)


eguajardo (approved):

move/its/sources/events.move
```diff
@@ -65,7 +65,11 @@ public(package) fun interchain_transfer<T>(
     amount: u64,
     data: &vector<u8>,
 ) {
-    let data_hash = bytes32::new(address::from_bytes(keccak256(data)));
+    let data_hash = if (data.length() == 0) {
```
milapsheth, line 68: test event in unit test

milapsheth: as part of reviewing tests, can you add event tests as well?

Foivos: added some tests

move/its/sources/events.move
```diff
@@ -128,3 +128,50 @@ public(package) fun unregistered_coin_received<T>(
...
     });
 }
+ 
+// ---------
+// Test Only 
+// ---------
+#[test_only]
+use its::coin::COIN;
+#[test_only]
+use its::token_id;
```
milapsheth, line 138:
```suggestion
#[test_only] {
    use its::coin::COIN;
    use its::token_id;
}
```
i wonder if Move supports this similar to Rust

move/its/sources/events.move
```diff
@@ -128,3 +128,50 @@ public(package) fun unregistered_coin_received<T>(
...
+
+// -----
+// Tests
+// -----
+#[test]
+fun test_interchain_transfer() {
```
milapsheth, line 144: This should be two separate tests
`test_interchain_transfer_event_empty_data`
`test_interchain_transfer_event_nonempty_data`

move/its/sources/events.move
```diff
@@ -128,3 +128,50 @@ public(package) fun unregistered_coin_received<T>(
...
+        destination_chain,
+        destination_address,
+        amount,
+        &data2,
+    );
+    let events = event::events_by_type<InterchainTransfer<COIN>>();
+
+    assert!(events.length() == 2);
+
+    assert!(events[0].data_hash == data1_hash);
```
milapsheth, line 175: check all event args

move/its/sources/its.move
```diff
@@ -417,6 +417,7 @@ fun test_register_coin() {
     let coin_management = its::coin_management::new_locked();
 
     register_coin(&mut its, coin_info, coin_management);
+    assert!(sui::event::events_by_type<its::events::CoinRegistered<COIN>>().length() == 1);
```
milapsheth, line 420:
```suggestion
    assert_event_len<its::events::CoinRegistered<COIN>>!(1);
```
create a macro fun `assert_event_len<$T>(len: u64)` in utils to make this readable and concise

milapsheth: we should add event checks in gateway tests as well if we don't have it

move/its/sources/events.move
```diff
@@ -128,3 +128,50 @@ public(package) fun unregistered_coin_received<T>(
...
+    assert!(events[1].data_hash == data2_hash);
```
milapsheth, line 176: should we add the event test on the actual interchain_transfer call, to ensure it's being called as part of the main call correctly?

milapsheth (approved):

move/its/sources/events.move
Foivos: it does not

move/its/sources/its.move
Foivos: added fun (not macro). Event querying in tests is new in move, will add tests in gateway in new PR

move/its/sources/events.move
Foivos: There is a test there too, (L514 in its.move) but it does not test for fields since they are not accessible outside the module.

move/its/sources/its.move
milapsheth: Yeah, separate PR for the gateway

move/its/sources/events.move
Foivos: done

milapsheth (approved):

PR #207, opened 2024-11-20
feat(axelar_gateway)!: check for events

Foivos:
[AXE-6678]

[AXE-6678]: https://axelarnetwork.atlassian.net/browse/AXE-6678?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/move/axelar_gateway/sources/gateway.move
+++ b/move/axelar_gateway/sources/gateway.move
@@ -603,7 +603,7 @@ fun test_take_approved_message() {
         payload,
     );
 
-    utils::assert_single_event<events::MessageExecuted>();
+    utils::assert_event<events::MessageExecuted>();
 
     let expected_approved_message = axelar_gateway::channel::create_approved_message(
         source_chain,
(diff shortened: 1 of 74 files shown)


move/axelar_gateway/sources/gateway.move
```diff
@@ -600,6 +602,9 @@ fun test_take_approved_message() {
         destination_id,
         payload,
     );
+
+    utils::single_event<events::MessageExecuted>();
```
milapsheth, line 606:
```suggestion
    assert_event<events::MessageExecuted>();
```
can you rename it to `assert_event` or `assert_emitted_event`? easier to understand

move/axelar_gateway/sources/gateway.move
```diff
@@ -777,6 +786,8 @@ fun test_rotate_signers() {
         ctx,
     );
 
+    utils::multiple_events<events::SignersRotated>(2);
```
milapsheth, line 789: this could be renamed to `assert_event_at` / `assert_event_at_index` maybe

milapsheth (approved):

move/axelar_gateway/sources/gateway.move
Foivos: it is about having the right number of evens in the test so far, but i could change the logic to only returning the last event and then that name would make more sense

PR #211, opened 2024-11-25
feat(its)!: add allow_function and disallow_function on ITS

Foivos:


move/its/sources/its.move
```diff
@@ -927,3 +940,31 @@ fun test_channel_address() {
...
+
+#[test]
+fun test_allow_function() {
+    let ctx = &mut sui::tx_context::dummy();
+    let mut self = create_for_testing(ctx);
+    let owner_cap = owner_cap::create(ctx);
+    let version = 0;
+    let function_name = b"function_name".to_ascii_string();
+
+    self.allow_function(&owner_cap, version, function_name);
```
milapsheth, line 952: this doesn't really test whether the function actually can be called or not. can you improve this in another PR?

milapsheth (approved):

move/its/sources/its.move
Foivos: we have tests for version control making sure that if a function is part of the allowed functions can be called, and tests ensuring that allow and disallow function adds/removes properly so what you are asking is implicitly tested, but can add that too

milapsheth (approved):

move/its/sources/its.move
milapsheth: We can do it more generically for all contracts in an integration test later maybe

PR #223, opened 2024-12-11
feat(its)!: rename its to interchain_token_service

Foivos:


move/its/sources/discovery.move
```diff
@@ -1,7 +1,7 @@
 module its::discovery;
 
 use abi::abi::{Self, AbiReader};
-use its::its::ITS;
+use its::interchain_token_service::InterchainTokenService;
```
milapsheth, line 4:
```suggestion
use interchain_token_service::interchain_token_service::InterchainTokenService;
```
the package should be renamed as well

test/testdata/interface_its_interchain_token_service.json
milapsheth, line 1: delete the old files after rename?

move/example/sources/its/its.move
```diff
@@ -8,7 +8,7 @@ use gas_service::gas_service::GasService;
 use its::coin_info;
 use its::coin_management;
 use its::discovery as its_discovery;
```
milapsheth, line 10: add changeset. also need to update deployment script

milapsheth (approved):

PR #225, opened 2024-12-12
feat(axelar_gateway)!: upgrade versioned

Foivos:


move/axelar_gateway/sources/gateway.move
```diff
@@ -174,6 +175,19 @@ entry fun rotate_signers(
     )
 }
 
+/// This function should only be called once
+/// (checks should be made on versioned to ensure this)
+/// It upgrades the version control to the new version control.
+entry fun migrate(self: &mut Gateway) {
+    let (v0, cap) = self.inner.remove_value_for_upgrade<Gateway_v0>();
+    let v1 = v0.migrate(version_control());
```
milapsheth, line 183: make the migration more interesting. e.g. add some message approval, and maybe another field to the gateway object, and add a check post upgrade?

move/axelar_gateway/sources/gateway.move
```diff
@@ -276,6 +290,7 @@ public fun take_approved_message(
 
 fun version_control(): VersionControl {
     version_control::new(vector[
+        vector[],
```
milapsheth, line 293: keep some functions available to test different cases

move/axelar_gateway/sources/gateway.move
```diff
@@ -174,6 +175,19 @@ entry fun rotate_signers(
     )
 }
 
+/// This function should only be called once
+/// (checks should be made on versioned to ensure this)
+/// It upgrades the version control to the new version control.
+entry fun migrate(self: &mut Gateway) {
```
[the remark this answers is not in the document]
Foivos: Since this is for testing I wanted to keep it simple, but will do

move/axelar_gateway/sources/gateway.move
milapsheth: also from the doc,

For the second upgrade, also include a change to a dependency contract, such as Utils or so. The deployment scripts should handle upgrading the dependency as well on a change, instead of reusing the existing config address. Can we also include upgrading the example contract to point to the new gateway contract?

Foivos: what is "some message approval"? Messages can be approved already. Can you be more specific? What check post upgrade?

move/axelar_gateway/sources/gateway.move
Foivos: cannot

move/axelar_gateway/sources/gateway.move
Foivos: since tests for allowed functions cannot be actually automated, I would have to write a custom function that lives on deployments that does nothing for our actual gateway to be able to test this 'new' field, how do you propose we go about this?

move/axelar_gateway/sources/gateway.move
```diff
@@ -174,6 +175,19 @@ entry fun rotate_signers(
     )
 }
 
+/// This function should only be called once
```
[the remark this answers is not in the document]
Foivos: you could call in many times in the same version, but it would still fail as there are checks

PR #236, opened 2025-01-21
feat(its)!: add chain name to token id derivation

Foivos:
[AXE-7210)]


move/interchain_token_service/sources/types/token_id.move
```diff
@@ -42,6 +43,7 @@ public fun from_info<T>(
 ): TokenId {
     let mut vec = address::from_u256(PREFIX_SUI_TOKEN_ID).to_bytes();
     vec.append(bcs::to_bytes(&type_name::get<T>()));
+    vec.append(bcs::to_bytes(chain_name));
```
milapsheth, line 46: add chain name before type name for consistent pattern

milapsheth, line 46: I think for consistency, we should save chain_name_hash on setup and use that instead

milapsheth: Since Sui packages are compiled at deployment-time, another option is to specify the chain name hash as an address during deployment. Seems like a bad pattern for this use case though haha

Foivos: as I mentioned earlier, chain_name_hash is not more efficient on Sui since length prefixing is 1 instead of 32 bytes (for reasonably sized strings at least)

milapsheth (approved):

move/interchain_token_service/sources/types/token_id.move
Foivos: done

PR #243, opened 2025-02-10
feat(gas-service)!: added the ability to accept any gas token

Foivos:
[AXE-4068]

[AXE-4068]: https://axelarnetwork.atlassian.net/browse/AXE-4068?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


move/gas_service/sources/versioned/gas_service_v0.move
```diff
@@ -123,18 +133,18 @@ module gas_service::gas_service_v0 {
     }
 
     #[test_only]
-    public(package) fun balance(self: &GasService_v0): &Balance<SUI> {
-        &self.balance
+    public(package) fun balance<T>(self: &GasService_v0): &Balance<T> {
```
milapsheth, line 136: Let's make this getter public since it'll be useful now with the switch to a Bag especially?

PR #268, opened 2025-06-23
feat(its): freeze coin metadata and migrate stored coin metadata 

drewstaylor:
[AXE-9693](https://axelarnetwork.atlassian.net/browse/AXE-9693)

[AXE-9693]: https://axelarnetwork.atlassian.net/browse/AXE-9693?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/move/interchain_token_service/sources/versioned/interchain_token_service_v0.move
+++ b/move/interchain_token_service/sources/versioned/interchain_token_service_v0.move
@@ -311,22 +311,21 @@ module interchain_token_service::interchain_token_service_v0 {
         let coin_data = self.coin_data<T>(token_id);
         let coin_info = coin_data.coin_info();
         let coin_management = coin_data.coin_management();
-        let has_metadata = option::is_some(coin_info.metadata());
 
         // TODO: fix needing to check both metadata (with / without)
-        let derived_token_id_with_metadata = token_id::from_coin_data(
-            &self.chain_name_hash,
-            coin_info,
-            coin_management,
-            has_metadata,
-        );
         let derived_token_id_without_metadata = token_id::from_coin_data(
             &self.chain_name_hash,
             coin_info,
             coin_management,
-            !has_metadata,
+            false,
         );
-        assert!((token_id == derived_token_id_with_metadata) || (token_id == derived_token_id_without_metadata), ENotCannonicalToken);
+        let derived_token_id_with_metadata = token_id::from_coin_data(
+            &self.chain_name_hash,
+            coin_info,
+            coin_management,
+            true,
+        );
+        assert!((token_id == derived_token_id_without_metadata) || (token_id == derived_token_id_with_metadata), ENotCannonicalToken);
 
         let name = coin_info.name();
         let symbol = coin_info.symbol();


move/interchain_token_service/sources/versioned/interchain_token_service_v0.move
```diff
@@ -653,10 +653,14 @@ module interchain_token_service::interchain_token_service_v0 {
...
 
+    // XXX TODO: version control?
+    public(package) fun migrate_coin<T>(self: &mut InterchainTokenService_v0, id: address) {
+        self.migrate_coin_metadata<T>(id);
+    }
+
     // -----------------
     // Private Functions
     // -----------------
-
```
milapsheth, line 659: keep the newline?

move/interchain_token_service/sources/interchain_token_service.move
```diff
@@ -103,6 +103,10 @@ module interchain_token_service::interchain_token_service {
         self.value_mut!(b"disallow_function").disallow_function(version, function_name);
     }
 
+    entry fun migrate_coin<T>(self: &mut InterchainTokenService, _: &OwnerCap, token_address: address) {
```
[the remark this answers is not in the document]
milapsheth: oh, is this because entry functions in Move can't use non-primitive types? Should still call the arg `token_id` for clarity

move/interchain_token_service/sources/versioned/interchain_token_service_v0.move
```diff
@@ -758,6 +762,18 @@ module interchain_token_service::interchain_token_service_v0 {
         );
     }
 
+    fun migrate_coin_metadata<T>(self: &mut InterchainTokenService_v0, id: address) {
+        let migration_token_id = token_id::from_address(id);
```
milapsheth, line 766:
```suggestion
    fun migrate_coin_metadata<T>(self: &mut InterchainTokenService_v0, token_id: address) {
        let token_id = token_id::from_address(token_id);
```
similar to the rust pattern of shadowing

move/interchain_token_service/sources/types/coin_info.move
```diff
@@ -21,14 +25,27 @@ module interchain_token_service::coin_info {
...
+        coin_info
+    }
+
+    // -----------------
+    // Package Functions
+    // -----------------
+
+    /// Publicly freeze metadata for a coin from the given `CoinInfo<T>`
+    public(package) fun release_metadata<T>(coin_info: &mut CoinInfo<T>) {
+        assert!(coin_info.metadata.is_some(), EMetadataEmpty);
```
milapsheth, line 46: for simplicity of operations, make this a no-op if the metadata was already none. Don't want one token_id to cause the whole batch to fail, and we don't need to filter by token_id type beforehand this way

move/interchain_token_service/sources/types/coin_info.move
```diff
@@ -85,8 +109,39 @@ module interchain_token_service::coin_info {
         assert!(coin_info.name() == name);
```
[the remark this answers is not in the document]
drewstaylor: This value gets moved, so we need the ref

move/interchain_token_service/sources/types/coin_info.move
```diff
@@ -21,14 +25,27 @@ module interchain_token_service::coin_info {
         }
     }
 
-    /// Create a new coin info from the given `CoinMetadata` object.
+    /// Create a new coin info from the given `CoinMetadata` object and publicly freeze the metadata object.
     public fun from_metadata<T>(metadata: CoinMetadata<T>): CoinInfo<T> {
```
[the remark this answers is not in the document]
milapsheth: yeah, the latter. Let's address in a separate PR

move/interchain_token_service/sources/types/coin_info.move
```diff
@@ -21,13 +22,27 @@ module interchain_token_service::coin_info {
...
-        CoinInfo {
+        let coin_info = CoinInfo {
             name: metadata.get_name(),
             symbol: metadata.get_symbol(),
             decimals: metadata.get_decimals(),
-            metadata: option::some(metadata),
+            metadata: option::none(),
+        };
+        transfer::public_freeze_object(metadata);
```
milapsheth, line 33: call release_metadata in the caller instead of doing this in a constructor

drewstaylor: I think adding this would have unintended behaviour as `CoinMetadata` is owned by the function at this point in the call stack and needs to be used somehow or destroyed by the function, because it's og purpose was to store the value. Since we can't change the signature or return type, and only <T> has drop, the `CoinMetadata` would get lost.

move/interchain_token_service/sources/interchain_token_service.move
```diff
@@ -103,6 +103,10 @@ module interchain_token_service::interchain_token_service {
         self.value_mut!(b"disallow_function").disallow_function(version, function_name);
     }
 
+    entry fun migrate_coin<T>(self: &mut InterchainTokenService, _: &OperatorCap, token_id: address) {
```
milapsheth, line 106:
```suggestion
    entry fun migrate_coin_metadata<T>(self: &mut InterchainTokenService, _: &OperatorCap, token_id: address) {
```
Perhaps, and add a doc string about what this does

move/interchain_token_service/sources/types/coin_info.move
milapsheth: Yeah, you're right, makes sense to freeze it here since we can't remove this method

milapsheth (approved):

move/interchain_token_service/sources/interchain_token_service.move
```diff
@@ -103,15 +103,45 @@ module interchain_token_service::interchain_token_service {
         self.value_mut!(b"disallow_function").disallow_function(version, function_name);
     }
 
+    /// Publicly freeze `CoinMetadata` for a token id and remove it from `CoinInfo` storage.
+    /// Can be used to migrate an existing coin from v0 to v1
```
milapsheth, line 107:
```suggestion
    /// Needs to be called after the v0 -> v1 upgrade
```

move/interchain_token_service/sources/interchain_token_service.move
```diff
@@ -103,15 +103,45 @@ module interchain_token_service::interchain_token_service {
...
+    entry fun migrate_coin_metadata<T>(self: &mut InterchainTokenService, _: &OperatorCap, token_id: address) {
+        self.value_mut!(b"migrate_coin_metadata").migrate_coin_metadata<T>(token_id);
+    }
+
     // ----------------
     // Public Functions
     // ----------------
+
+    // TODO: disable this in version control
```
milapsheth, line 116: Let's just remove it from the v0 list in this PR

move/interchain_token_service/sources/interchain_token_service.move
```diff
@@ -103,15 +103,45 @@ module interchain_token_service::interchain_token_service {
...
     public fun register_coin<T>(self: &mut InterchainTokenService, coin_info: CoinInfo<T>, coin_management: CoinManagement<T>): TokenId {
         let value = self.value_mut!(b"register_coin");
 
         value.register_coin(coin_info, coin_management)
```
milapsheth, line 120: Make it a one-liner perhaps to match other usage

milapsheth, line 120: Actually, since we don't want people to use this method from v1, let's just abort with not supported error and remove the internal call? Tests should use the other two methods

move/interchain_token_service/sources/interchain_token_service.move
```diff
@@ -103,15 +103,45 @@ module interchain_token_service::interchain_token_service {
...
+        symbol: ascii::String,
+        decimals: u8,
+        coin_management: CoinManagement<T>,
+    ): TokenId {
+        let value = self.value_mut!(b"register_coin_from_info");
+
+        value.register_coin_from_info(name, symbol, decimals, coin_management)
+    }
+
+    public fun register_coin_from_metadata<T>(
```
milapsheth, line 135: Add dcostrings on usage

move/interchain_token_service/sources/interchain_token_service.move
```diff
@@ -464,6 +494,8 @@ module interchain_token_service::interchain_token_service {
             // Version 1
             vector[
                 b"register_coin",
```
milapsheth, line 496: We can remove this since it shouldn't be called

move/interchain_token_service/sources/versioned/interchain_token_service_v0.move
```diff
@@ -181,6 +181,36 @@ module interchain_token_service::interchain_token_service_v0 {
         coin_management: CoinManagement<T>,
     ): TokenId {
         let token_id = token_id::from_coin_data(&self.chain_name_hash, &coin_info, &coin_management);
+        let mut coin_data = coin_data::new(coin_management, coin_info);
+        let coin_info = coin_data.coin_info_mut();
+
+        // This will no-op if coin_info.metadata is None
+        coin_info.release_metadata();
```
milapsheth, line 188: This is not needed since we're disabling this method

move/interchain_token_service/sources/versioned/interchain_token_service_v0.move
```diff
@@ -181,6 +181,36 @@ module interchain_token_service::interchain_token_service_v0 {
...
+
+    public(package) fun register_coin_from_metadata<T>(
+        self: &mut InterchainTokenService_v0,
+        metadata: &CoinMetadata<T>,
+        coin_management: CoinManagement<T>,
+    ): TokenId {
+        let coin_info = coin_info::from_metadata_ref<T>(metadata);
+        let token_id = token_id::from_coin_data(&self.chain_name_hash, &coin_info, &coin_management);
 
         self.add_registered_coin(token_id, coin_data::new(coin_management, coin_info));
```
milapsheth, line 215: Call register_coin instead of duplicating the logic. Pass a new bool arg to register_coin that indicates whether coin metadata was used or not to create coin info. Use this bool in the token derivation

move/interchain_token_service/sources/versioned/interchain_token_service_v0.move
```diff
@@ -181,6 +181,36 @@ module interchain_token_service::interchain_token_service_v0 {
         coin_management: CoinManagement<T>,
     ): TokenId {
         let token_id = token_id::from_coin_data(&self.chain_name_hash, &coin_info, &coin_management);
```
milapsheth, line 183: This needs to be changed to pass in a bool. We can't rely on checking if coin metadata is stored in coin info anymore for the derivation

move/interchain_token_service/sources/interchain_token_service.move
milapsheth: yeah, need to update tests to use the new methods

drewstaylor:
Seems like some JS tests will need updating after these changes (re: `test-js` failures). Assessing and tracing the errors now.

move/interchain_token_service/sources/interchain_token_service.move
drewstaylor: All the `entry` functions are one liner, and all the `public` functions are these two-liners. Do you prefer I streamline everything to one-liners?

drewstaylor: Created [9747](https://axelarnetwork.atlassian.net/browse/AXE-9747) for this

move/interchain_token_service/sources/interchain_token_service.move
```diff
@@ -118,13 +118,18 @@ module interchain_token_service::interchain_token_service {
...
     #[allow(dead_code)]
     public fun register_coin<T>(self: &mut InterchainTokenService, coin_info: CoinInfo<T>, coin_management: CoinManagement<T>): TokenId {
         abort EUnsupported;
         let value = self.value_mut!(b"register_coin");
-        value.register_coin(coin_info, coin_management)
+        value.register_coin(coin_info, coin_management, option::is_some(coin_info.metadata()))
     }
 
+    /// Register a coin from user supplied values. Replaces legacy function `register_coin`.
+    /// @since version 1.0
```
milapsheth, line 132:
```suggestion
```
don't think we need to add comments for this, we should rely on version control instead. of npm package version is different from the internal version tag in the contract, so the label here can be confusing

move/interchain_token_service/sources/interchain_token_service.move
```diff
@@ -118,13 +118,18 @@ module interchain_token_service::interchain_token_service {
     // ----------------
     // Public Functions
     // ----------------
+
+    /// Legacy function to register a coin from the given `CoinInfo`.
+    /// @deprecated since version 1.0
```
milapsheth, line 123:
```suggestion
    /// @deprecated
```
version tag can be confusing since there are different ones we have

move/interchain_token_service/sources/interchain_token_service.move
```diff
@@ -137,6 +142,8 @@ module interchain_token_service::interchain_token_service {
         value.register_coin_from_info(name, symbol, decimals, coin_management)
     }
 
+    /// Register a coin from the given `CoinMetadata` reference. Replaces legacy function `register_coin`.
+    /// @since version 1.0
```
milapsheth, line 146:
```suggestion
```

move/interchain_token_service/sources/versioned/interchain_token_service_v0.move
```diff
@@ -175,19 +175,29 @@ module interchain_token_service::interchain_token_service_v0 {
         &self.version_control
     }
 
+    /// Originally, the versioned function for public fun interchain_token_service::register_coin.
+    /// Now, a worker function to process registering a coin (modified in version 1.0).
+    /// @see self.register_coin_from_info
+    /// @see self.register_coin_from_metadata
```
milapsheth, line 181:
```suggestion
```
Not too useful I think, should be clear by following references from the code itself. Ideally, comments should only be added if the code can't be made self-descriptive, since they add maintenance overhead

move/interchain_token_service/sources/versioned/interchain_token_service_v0.move
```diff
@@ -305,10 +315,22 @@ module interchain_token_service::interchain_token_service_v0 {
...
+            coin_management,
+            has_metadata,
+        );
+        let derived_token_id_b = token_id::from_coin_data(
+            &self.chain_name_hash,
+            coin_info,
+            coin_management,
+            !has_metadata,
+        );
+        assert!((token_id == derived_token_id_a) || (token_id == derived_token_id_b), ENotCannonicalToken);
```
milapsheth, line 333:
```suggestion
        // TODO: fix needing to check both metadata (with / without)
        let derived_token_id_without_metadata = token_id::from_coin_data(
            &self.chain_name_hash,
            coin_info,
            coin_management,
            false,
        );
        let derived_token_id_with_metadata = token_id::from_coin_data(
            &self.chain_name_hash,
            coin_info,
            coin_management,
            true,
        );
        assert!((token_id == derived_token_id_without_metadata) || (token_id == derived_token_id_with_metadata), ENotCannonicalToken);
```
coin info won't have metadata anymore. and let's use more meaningful names

milapsheth (approved):

PR #270, opened 2025-07-02
fix(its): deprecated register_coin should release coin metadata

drewstaylor:
[AXE-9770](https://axelarnetwork.atlassian.net/browse/AXE-9770)

[AXE-9770]: https://axelarnetwork.atlassian.net/browse/AXE-9770?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/move/interchain_token_service/sources/versioned/interchain_token_service_v0.move
+++ b/move/interchain_token_service/sources/versioned/interchain_token_service_v0.move
@@ -176,7 +176,7 @@ module interchain_token_service::interchain_token_service_v0 {
     }
 
     /// Legacy function to register a coin from the given `CoinInfo`, provides backwards compatibility.
-    /// Freezing the metadata in legacy function prevents the coin's metadata from needing to be migrated 
+    /// Freezing the metadata in legacy function prevents the coin's metadata from needing to be migrated
     /// later with migrate_coin_metadata
     /// @deprecated
     public(package) fun register_coin<T>(


drewstaylor:
> ### Bug: Inconsistent Metadata Handling in Coin Registration
> The deprecated `register_coin` function calls `coin_info.release_metadata()` when registering coins with metadata. However, the `register_coin_from_metadata` function, which also handles metadata, omits this call. This inconsistency leads to different metadata handling and potential storage states for registered coins depending on the registration path used.
> 
> `move/interchain_token_service/sources/versioned/interchain_token_service_v0.move#L191-L193`
> https://github.com/axelarnetwork/axelar-cgp-sui/blob/31f8972bce243e19835d9683e7f6b5d626b75c7e/move/interchain_token_service/sources/versioned/interchain_token_service_v0.move#L191-L193
> 
> `move/interchain_token_service/sources/versioned/interchain_token_service_v0.move#L241-L251`
> [Fix in Cursor](https://cursor.com/open?data=eyJhbGciOiJSUzI1NiIsInR5cCI6IkpXVCIsImtpZCI6ImJ1Z2JvdC12MSJ9.eyJ2ZXJzaW9uIjoxLCJ0eXBlIjoiQlVHQk9UX0ZJWF9JTl9DVVJTT1IiLCJkYXRhIjp7InJlZGlzS2V5IjoiYnVnYm90OjU2YzM3NTIxLTJkNDQtNDYzOC04Nzc4LWZmYzU4Yjk0ODA0ZSIsImVuY3J5cHRpb25LZXkiOiJ4aDk4RWU5UVFYVEJPR0dvTjlJRTZaRWIweGNnRjVmY1dFLUJZZUswc2JnIiwiYnJhbmNoIjoiZmVhdC9yZWdpc3Rlci1jb2lucy1iYWNrd2FyZHMtY29tcGF0YWJpbGl0eSJ9LCJpYXQiOjE3NTE0ODYzNDQsImV4cCI6MTc1MjA5MTE0NH0.c15fXru6cqKWDj-Mz-NjqUfaQ37AfwNA0OKc9Q7okRzNOw4qa1CXD00y42dv7ltmbG92G46HFMoZmIvGbEMI5aw8G8O2Pyox8ERm8W6Ry1CNKbGet7waRv3sCq3_AgwKKha-ja-ilG0nyA6IUJwCJek6uvwYf7xkpKv-qIDFqOISjV8gOdFJEZa1cINNBZP7k6zrlOrpUR5_DigjvJ8NrOuT3t3egLzPxyMhKfyevaqKz6Yg_haLfMXejW_-Hpmmaj5JDJLWFAqEABPZTD1AyOAc-0rqSFlFSjsLdP6uvQIXA5dTrFbX2_CrQZYX0UIZxPsAPnhlRpzsMEVjb9TDzg)
> 

This is the intended behaviour, because `register_coin_from_metadata` doesn't store the metadata in `CoinInfo` but the deprecated `register_coin` does. Freezing the metadata in legacy function prevents the coin's metadata from needing to be migrated later with `migrate_coin_metadata`

move/interchain_token_service/sources/versioned/interchain_token_service_v0.move
```diff
@@ -215,7 +239,16 @@ module interchain_token_service::interchain_token_service_v0 {
     ): TokenId {
         let coin_info = coin_info::from_metadata_ref<T>(metadata);
 
-        self.register_coin(coin_info, coin_management, true)
+        let token_id = token_id::from_coin_data(
```
milapsheth, line 242: i don't like the duplication. Introduce a new `register_coin_from_coin_info` that's deprecated, releases the metadata, and calls the new `register_coin` helper

move/interchain_token_service/sources/versioned/interchain_token_service_v0.move
```diff
@@ -175,22 +175,37 @@ module interchain_token_service::interchain_token_service_v0 {
         &self.version_control
     }
 
+    /// Legacy function to register a coin from the given `CoinInfo`.
```
milapsheth, line 178: Explain why we're doing this. If BugBot doesn't have enough context to understand why this is intentional, then the code/comments aren't descriptive enough :p

move/interchain_token_service/sources/interchain_token_service.move
```diff
@@ -24,12 +24,6 @@ module interchain_token_service::interchain_token_service {
     const VERSION: u64 = 1;
```
milapsheth, line 24: PR title is vague, which function, what compatibility? Maybe something like
`fix(its): deprecated register_coin should release coin metadata`

move/interchain_token_service/sources/versioned/interchain_token_service_v0.move
drewstaylor: `register_coin` should be the function name in `interchain_token_service_v0` because, even though they are no longer versioned, that's what the `entry fun` is called in `interchain_token_service`. However, I'll remove the duplication

drewstaylor:
> ### Bug: Test Fails Due to Backward Compatibility
> The `test_deprecated_register_coin_should_fail` test incorrectly expects the deprecated `register_coin` function to fail with `EFunctionNotSupported`. The `register_coin` function is implemented to work for backward compatibility by calling its underlying v0 implementation. Additionally, the `create_for_testing` setup allows all functions via a wildcard, preventing the expected failure. The test should either expect success and verify correct behavior, or its setup should be adjusted to properly restrict the deprecated function.

> ### Bug: Deprecated Function Excluded from Version Control
> The deprecated `register_coin` function's implementation is unreachable. Although its logic was recently changed for backwards compatibility, calls to it will always fail with `EFunctionNotSupported` because it is not included in the `version_control` allowed functions list for any version. This makes the new implementation dead code, as confirmed by `test_deprecated_register_coin_should_fail`.

For clarity:
* The `interchain_token_service::register_coin` function fails with the expected error as it's excluded from version control. 
* `interchain_token_service_v0::register_coin` is the backwards compatible one

PR #272, opened 2025-07-21
feat(its): migrate function for version control

drewstaylor:
This PR adds a migrate entry point function to ITS, and its corresponding package function to its_v0. Additionally it adds unit test coverage.

The code I used was borrowed from [this branch](https://github.com/axelarnetwork/axelar-cgp-sui/tree/feat/upgrade-testing) but I think we should find a different way for enforcing linear upgrades and avoiding migrating to the same upgrade twice, because it blocks unit testing coverage. The `ECannotMigrateTwice` assertion is currently commented out.

### DIFF
--- a/move/interchain_token_service/sources/interchain_token_service.move
+++ b/move/interchain_token_service/sources/interchain_token_service.move
@@ -118,8 +118,8 @@ module interchain_token_service::interchain_token_service {
     /// This function should only be called once
     /// (checks should be made on versioned to ensure this)
     /// It upgrades the version control to the new version control.
-    entry fun migrate(self: &mut InterchainTokenService, _: &OwnerCap, data: vector<u8>) {
-        self.inner.load_value_mut<InterchainTokenService_v0>().migrate(version_control(), data);
+    entry fun migrate(self: &mut InterchainTokenService, _: &OwnerCap) {
+        self.inner.load_value_mut<InterchainTokenService_v0>().migrate(version_control());
     }
 
     // ----------------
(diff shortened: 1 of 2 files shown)


move/interchain_token_service/sources/versioned/interchain_token_service_v0.move
```diff
@@ -119,6 +123,18 @@ module interchain_token_service::interchain_token_service_v0 {
         }
     }
 
+    public(package) fun version_control(self: &InterchainTokenService_v0): &VersionControl {
+        &self.version_control
+    }
+
+    public(package) fun migrate(self: &mut InterchainTokenService_v0, version_control: VersionControl, data: vector<u8>) {
+        // TODO: enforcing `ECannotMigrateTwice` assertion will block unit tests
```
Foivos, line 131: You should add this indeed.

move/interchain_token_service/sources/interchain_token_service.move
```diff
@@ -115,6 +115,13 @@ module interchain_token_service::interchain_token_service {
         self.value_mut!(b"migrate_coin_metadata").migrate_coin_metadata<T>(token_id);
     }
 
+    /// This function should only be called once
+    /// (checks should be made on versioned to ensure this)
+    /// It upgrades the version control to the new version control.
+    entry fun migrate(self: &mut InterchainTokenService, _: &OwnerCap, data: vector<u8>) {
```
Foivos, line 121: Since entry functions can be changed in the future and we do not plan to use data I propose you remove data.

PR #277, opened 2025-08-18
fix(scripts): retry fetching transactions if `objectChanges` field is undefined

jcs47:


src/common/tx-builder-base.ts
```diff
@@ -249,7 +249,7 @@ export class TxBuilderBase {
             },
         });
 
-        if (!result.confirmedLocalExecution) {
+        if (!result.confirmedLocalExecution || !result.objectChanges) {
```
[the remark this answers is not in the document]
Foivos: First of this if is not inside a loop, so what it's saying doesn't make sense. Second of all that array can never be empty because at least the gas object will be modified. Third, the thing that the RPC returns should always be properly populated, but we could check for errors with the output.

Foivos (approved):

src/common/tx-builder-base.ts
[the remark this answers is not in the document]
jcs47: but I have tested the script with this retry, and what I found is that it is possible for the tx to be returned without any object changes, but then upon fetching the transaction again, the `objectChanges` field is finally populated. I don't know if this is expected behaviour, but we need to account for this possibility anyway.

The only question is, is there any scenario where the `objectChanges` field does not get populated, or can we safely have the script retry until the field is populated??

src/common/tx-builder-base.ts
```diff
@@ -249,26 +253,42 @@ export class TxBuilderBase {
...
-                            showObjectChanges: true,
-                            ...options,
-                        },
-                    });
-                    break;
-                } catch (e) {
-                    console.log(e);
-                    await new Promise((resolve) => setTimeout(resolve, 1000));
-                }
+        const maxRetries = 10; // 10 seconds with 1-second delay
```
milapsheth, line 256:
```suggestion
        const maxRetries = 10;
        const delay = 1000;  // 1 second
```

src/common/tx-builder-base.ts
```diff
@@ -249,26 +253,42 @@ export class TxBuilderBase {
...
+                    digest: result.digest,
+                    options: {
+                        showEffects: true,
+                        showObjectChanges: true,
+                        ...options,
+                    },
+                });
+            } catch (e) {
+                console.log(e);
+                await new Promise((resolve) => setTimeout(resolve, 1000));
```
milapsheth, line 278:
```suggestion
                await new Promise((resolve) => setTimeout(resolve, delay));
```

src/common/tx-builder-base.ts
```diff
@@ -249,26 +253,42 @@ export class TxBuilderBase {
...
-                } catch (e) {
-                    console.log(e);
-                    await new Promise((resolve) => setTimeout(resolve, 1000));
-                }
+        const maxRetries = 10; // 10 seconds with 1-second delay
+        let retries = 0;
+
+        while (true) {
+
+            console.log(`${result.confirmedLocalExecution} ${expectObjChanges} ${result.objectChanges}`);
```
milapsheth, line 261: avoid logging in helpers outside of debugging sessions

src/common/tx-builder-base.ts
```diff
@@ -249,26 +253,42 @@ export class TxBuilderBase {
...
-                    });
-                    break;
-                } catch (e) {
-                    console.log(e);
-                    await new Promise((resolve) => setTimeout(resolve, 1000));
-                }
+        const maxRetries = 10; // 10 seconds with 1-second delay
+        let retries = 0;
+
+        while (true) {
```
milapsheth, line 259:
```suggestion
        while (retries < maxRetries) {
```
check max retry in the loop and throw error at the end of the function instead?

src/common/tx-builder-base.ts
```diff
@@ -249,26 +253,42 @@ export class TxBuilderBase {
...
+                result = await this.client.getTransactionBlock({
+                    digest: result.digest,
+                    options: {
+                        showEffects: true,
+                        showObjectChanges: true,
+                        ...options,
+                    },
+                });
+            } catch (e) {
+                console.log(e);
```
milapsheth, line 277: avoid debug logging

src/common/tx-builder-base.ts
milapsheth, line 256: you can release a snapshot from this branch via the gh action and update deployments repo against it to test it out fyi

milapsheth (approved):

src/common/tx-builder-base.ts
```diff
@@ -229,7 +229,11 @@ export class TxBuilderBase {
         });
     }
```
milapsheth, line 231: run `npm run cs` to create a `patch` changelog entry for this

src/common/tx-builder-base.ts
jcs47: this one was already part of the code

PR #285, opened 2025-09-17
feat(its)!: fix everything in relayer discovery

Foivos:


drewstaylor (approved):
Testnet on devnet amplifier and confirmed to be working

PR #286, opened 2025-09-22
fix: update move toml and lock files correctly based on version and network

drewstaylor:
This PR includes the following changes:

* deprecated faucet function`requestSuiFromFaucetV0` is updated to `requestSuiFromFaucetV2` (with fallback to v0 for `localnet`)
* deprecated type alias `TomlPrimitive` is updated to `TomlValue`
* function `updateMoveToml` now updates move toml and lock files correctly for upgraded contracts, based on three new param inputs: version, sui network and legacy package id

### DIFF
--- a/.changeset/easy-llamas-eat.md
+++ b/.changeset/easy-llamas-eat.md
@@ -1,0 +1,7 @@
+---
+'@axelar-network/axelar-cgp-sui': patch
+---
+
+- deprecated faucet function V0 is updated to V2 (with fallback to V0 for localnet)
+- deprecated type alias TomlPrimitive is updated to TomlValue
+- updateMoveToml now updates move toml and lock files correctly for upgraded contracts
--- a/.changeset/easy-llamas-eat.md
+++ b/.changeset/easy-llamas-eat.md
@@ -2,6 +2,4 @@
 '@axelar-network/axelar-cgp-sui': patch
 ---
 
-- deprecated faucet function V0 is updated to V2 (with fallback to V0 for localnet)
-- deprecated type alias TomlPrimitive is updated to TomlValue
-- updateMoveToml now updates move toml and lock files correctly for upgraded contracts
+updateMoveToml now updates move toml and lock files correctly for upgraded contracts
--- a/src/common/utils.ts
+++ b/src/common/utils.ts
@@ -17,6 +17,10 @@ export const fundAccountsFromFaucet = async (addresses: string[]) => {
                 });
             }
 
+            case 'mainnet': {
+                throw new Error(`Faucet request failed, invalid network: ${network}`);
+            }
+
             default: {
                 return requestSuiFromFaucetV2({
                     host: getFaucetHost(network as 'devnet' | 'testnet'),
--- a/src/common/utils.ts
+++ b/src/common/utils.ts
@@ -16,6 +16,7 @@ export const fundAccountsFromFaucet = async (addresses: string[]) => {
                     recipient: address,
                 });
             }
+
             default: {
                 return requestSuiFromFaucetV2({
                     host: getFaucetHost(network as 'devnet' | 'testnet'),
--- a/src/common/utils.ts
+++ b/src/common/utils.ts
@@ -1,5 +1,5 @@
 import { getFullnodeUrl, SuiMoveNormalizedType } from '@mysten/sui/client';
-import { getFaucetHost, requestSuiFromFaucetV2 } from '@mysten/sui/faucet';
+import { getFaucetHost, requestSuiFromFaucetV0, requestSuiFromFaucetV2 } from '@mysten/sui/faucet';
 import { arrayify, keccak256 } from 'ethers/lib/utils';
 import secp256k1 from 'secp256k1';
 import { STD_PACKAGE_ID } from './types';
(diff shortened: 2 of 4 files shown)


src/common/utils.ts
```diff
@@ -8,10 +8,22 @@ export const fundAccountsFromFaucet = async (addresses: string[]) => {
...
+                    recipient: address,
+                });
+            }
+
+            default: {
+                return requestSuiFromFaucetV2({
+                    host: getFaucetHost(network as 'devnet' | 'testnet'),
+                    recipient: address,
+                });
+            }
```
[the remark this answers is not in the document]
drewstaylor: Mainnet has no faucet

.changeset/easy-llamas-eat.md
```diff
@@ -0,0 +1,7 @@
+---
+'@axelar-network/axelar-cgp-sui': patch
+---
+
+- deprecated faucet function V0 is updated to V2 (with fallback to V0 for localnet)
+- deprecated type alias TomlPrimitive is updated to TomlValue
```
milapsheth, line 6: doesn't seem useful to an external user to include this in a changelog

src/node/node-utils.ts
```diff
@@ -80,34 +93,116 @@ export function updateMoveToml(
...
+    network?: undefined | string,
+    originalPackageId?: undefined | string,
 ) {
+    if (typeof version !== 'number') {
+        version = 0;
+    }
+
+    if (typeof network !== 'string') {
+        network = 'testnet';
+    }
```
milapsheth, line 108: why allow a default? require it?

src/node/node-utils.ts
```diff
@@ -80,34 +93,116 @@ export function updateMoveToml(
...
+        // [env]
+        if (!lockJson.env) {
+            lockJson.env = {};
+        }
+
+        if (network !== 'devnet' && network !== 'testnet' && network !== 'mainnet') {
+            throw new Error(
+                `Unrecognized chain-id for contract upgrade for given network ${network}. Must be one of ${JSON.stringify(chainIds)}`,
+            );
+        }
```
milapsheth, line 186: do such validation as early as possible, in this case near the top of the function

src/node/node-utils.ts
```diff
@@ -80,34 +93,116 @@ export function updateMoveToml(
...
-    (moveJson as Record<string, Record<string, string>>).addresses[packageName] = packageId;
+        // Version >= 1
+        // Remove the 'published-at' field (if required)
+        let legacyPackageId;
+
+        if (originalPackageId) {
+            legacyPackageId = originalPackageId;
+        } else if ((tomlJson as Record<string, Record<string, string>>).package['published-at']) {
+            legacyPackageId = (tomlJson as Record<string, Record<string, string>>).package['published-at'];
+            if (!legacyPackageId) throw new Error(`Upgrade parameter missing, no published-at id was found for given path: ${tomlPath}`);
```
milapsheth, line 153: use braces

src/node/node-utils.ts
```diff
@@ -80,34 +93,116 @@ export function updateMoveToml(
...
+        // Parse the Move.lock file as JSON
+        lockJson = toml.parse(lockRaw);
+
+        // If syncing an already published upgrade, use 'original-published-id' in place of 'published-at'
+        // for deriving the legacy package id
+        if (!legacyPackageId) {
+            try {
+                legacyPackageId = (lockJson as Record<string, Record<string, string>>)[`env.${network}`]['original-published-id'];
+            } catch {
+                throw new Error(`Upgrade parameter missing, no original-published-id was found for given path: ${lockPath}`);
```
milapsheth, line 172: won't this just be `undefined` and not raise an error if the key is missing. Instead set the value and test if it's set or throw an error

src/node/node-utils.ts
drewstaylor: if we require this parameter it's no longer backwards compatible

src/node/node-utils.ts
```diff
@@ -80,34 +93,116 @@ export function updateMoveToml(
...
+    } else {
+        if (!wasBuilt) {
+            throw new Error(`Upgrade parameter missing, no lock file was found for given path: ${lockPath}`);
+        }
 
-    // Update the package address under the addresses section e.g. gas_service = "0x1"
-    (moveJson as Record<string, Record<string, string>>).addresses[packageName] = packageId;
+        // Version >= 1
+        // Remove the 'published-at' field (if required)
+        let legacyPackageId;
```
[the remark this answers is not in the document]
drewstaylor: Probably, but we will need to re-do all tests if we make such a root level change because that's not what we've tested. If that's what you want I can do another dry-run tomorrow. Lmk.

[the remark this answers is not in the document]
milapsheth: We should simplify it given this is already deprecated and might cause us more problems later. Just need to do an upgrade test, we should be able to just copy/paste most of the cmds from the release doc ideally

src/node/node-utils.ts
```diff
@@ -2,9 +2,23 @@ import { execSync } from 'child_process';
 import fs from 'fs';
 import path from 'path';
 import { Bytes } from 'ethers';
-import toml, { TomlPrimitive } from 'smol-toml';
+import toml, { TomlValue } from 'smol-toml';
 import { Dependency, DependencyNode, InterchainTokenOptions } from '../common/types';
 
+type ChainType = {
```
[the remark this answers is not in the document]
drewstaylor: how would we know the chain id to put in the lock file?

src/node/node-utils.ts
```diff
@@ -2,9 +2,23 @@ import { execSync } from 'child_process';
...
+    devnet: string;
+    testnet: string;
+    mainnet: string;
+};
+
+const emptyPackageId = '0x0';
+
+const chainIds: ChainType = {
```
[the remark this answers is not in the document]
drewstaylor: I think it depends on the configuration of the local node, and whether they're using testnet or devnet build?

drewstaylor:
Before merging this I will test on-chain run through again

drewstaylor:
> Before merging this I will test on-chain run through again

Dry-run completed successfully, this PR is ready to be merged

milapsheth (approved):

